PR #6, opened 2026-09-05
fix: the Review rounds page's sort and filter survive the bundler

oleksandrdubyna88:
0.29.10's error trap reported it a minute after release: "rowMatches is not defined". The page
embeds compareRows and rowMatches by their source text so the tested function is the one that runs
— and the extension ships bundled and minified, where a function that is not a top-level export is
renamed: the page received `function m(a, b, c, d)` and called rowMatches(). The unbundled out/ in
node and in headless Chromium rendered every row, which is why nothing caught it before the VSIX.

Bound by assignment now (`var rowMatches = <source>`), which holds whatever the bundler calls the
function inside; a test refuses a bare declaration. Extension 0.29.11.

Co-Authored-By: Claude Fable 5.1 <noreply@anthropic.com>

## Summary by CodeRabbit

- **Новые возможности**
  - Обновлено отображение активных раундов: этап, номер, ветка, статус и счётчик gating теперь выводятся отдельными строками.
  - Длинные названия веток корректно переносятся в узкой панели.

- **Исправления**
  - Исправлена работа сортировки и фильтрации на странице Review rounds в установленной версии расширения.
  - Выпущена версия расширения 0.29.11.

### DIFF
--- a/src_vs_code/CHANGELOG.md
+++ b/src_vs_code/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Changelog
 
+## 0.29.11 — 2026-09-05 (server 0.17.3)
+
+**The Review rounds page draws its rows in the installed extension, not only on a developer's
+machine.** 0.29.10's error trap reported it within a minute of release: *rowMatches is not
+defined*. The page embeds its sort and filter functions by their source text so the tested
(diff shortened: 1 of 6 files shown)


oleksandrdubyna88:
@coderabbitai review

PR #10, opened 2026-09-05
fix: the registry URL escapes every slash of a package name

oleksandrdubyna88:
CodeQL `js/incomplete-sanitization` at cliVersions.ts:132 — `replace('/', …)` handled the first occurrence only; `replaceAll` now, with a test that pins the escaping. Closes the one open code-scanning alert on this repository.

### DIFF
--- a/.github/workflows/sonarcloud.yml
+++ b/.github/workflows/sonarcloud.yml
@@ -76,7 +76,11 @@ jobs:
           # with tests. c8 reads the same V8 data and applies the source maps, so the report names
           # the TypeScript. (Measured 2026-09-05 on PR #10: SF:out\cliVersions.js against
           # SF:src\cliVersions.ts.)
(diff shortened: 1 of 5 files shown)


oleksandrdubyna88:
Rebased on main, and the coverage gate is now measured against the TypeScript rather than the compiled output.

The gate failed here on `new_coverage 0 %` even though the change ships with a test that covers exactly the new function. The cause was the report, not the change: `node --test --experimental-test-coverage` writes the paths of the files it executed (`out/cliVersions.js`) and SonarCloud analyses the sources (`src/cliVersions.ts`), so no coverage was ever attributed to any TypeScript line. Every extension pull request has been arriving at this gate as 0 % for that reason.

`c8` reads the same V8 coverage data and applies the source maps; verified locally that the report goes from `SF:out\cliVersions.js` to `SF:src_vs_code/src/cliVersions.ts`.

PR #45, opened 2026-09-05
fix(release): the server ships the SQLite library it opens its database through

oleksandrdubyna88:
**The server shipped unable to open its own database — on every platform, for every user.**

0.18.1 carried the executable alone. Native AOT compiles managed code; the call into SQLite still resolves at run time through the OS loader, which searches the directory the executable sits in. So the installed server threw `DllNotFoundException` the first time anything touched the rounds database:

```
System.TypeInitializationException: A type initializer threw an exception.
 ---> System.DllNotFoundException: Unable to load DLL 'e_sqlite3' or one of its dependencies
```

That write is **best-effort by design** — a database that cannot be written must never take down a round somebody is waiting for — so it failed in silence. Rounds ran, findings were answered, nothing was recorded, and the log page had nothing to show. `--log` on a machine that *did* have a database crashed outright.

**Every check passed throughout**, which is the part worth keeping: `--version`, `--help` and a full `tools/list` exchange never touch SQLite, so the AOT publish check, the contract smoke and 751 tests all stayed green. It was found by installing the release and running a real round against it.

## Three fixes

1. **The archive carries the native library** — `e_sqlite3.dll`, `libe_sqlite3.so`, `libe_sqlite3.dylib`, whichever the RID produced — and the job **fails** when the publish output has none, rather than shipping a server that cannot open its own database.
2. **The smoke makes the published binary open one.** An empty file is a valid empty SQLite database, so `--log` against it loads the library and reads through it. That is the check that would have caught this, and it now runs for all six RIDs.
3. **The installer copies everything the archive brought**, not the single file it knows by name — so the next native dependency needs no change there. A companion that cannot be copied is not fatal: the binary is in place and a feature degrades rather than the server failing to start.

## Verified

- The failure reproduced on the installed 0.18.1 binary, then went away with the library placed beside it, then a real plan round wrote `coai.db` (90 KB) and `--log` read the round back with its four decisions and the findings' text.
- **440 tests pass.** The new one reads the release workflow and goes red when the native library leaves it.

## Summary by CodeRabbit

- **Исправления**
  - Исправлена поставка нативной библиотеки SQLite: архивы релиза теперь содержат необходимые файлы, а установщик копирует их рядом с сервером.
  - Установка сообщает об ошибке, если обязательная библиотека SQLite отсутствует или не скопирована.
  - Команда `--log` больше не завершается сбоем при недоступной SQLite-библиотеке: выводится объяснение и пустой журнал.
  - Исправлено отображение записей раундов на странице журнала.

- **Релиз**
  - Версия расширения обновлена до 0.30.4.


oleksandrdubyna88:
@coderabbitai review

oleksandrdubyna88:
All three findings are fixed in `e2e38a9` — thank you, and the first one was caused by this very branch.

1. **`release.yml` — stderr after a successful `--log`.** Exactly right, and it was self-inflicted: this branch made `--log` answer a missing library with an empty log, a note on stderr and exit 0, so the smoke added to catch that failure would have passed with the library missing. The smoke now greps stderr on success too. Not "stderr is empty", though: a run against an empty database legitimately prints `no such table: findings` and exits 0, which is SQLite answering *through* the loaded library and exactly what the smoke wants to see — measured before writing the condition. A test now reads the grepped sentence out of the workflow and asserts the server prints it, so the two files cannot drift apart.
2. **The requirement now comes from the RID, not from the archive.** Your argument is the stronger one: the release job's check protects future archives and can do nothing for 0.18.1, which is already published without the library. `requiredCompanionMissing` computes the platform's name (`e_sqlite3.dll` / `libe_sqlite3.so` / `libe_sqlite3.dylib`) and refuses when it is not in place. The test that asserted the old behaviour is rewritten with the reasoning that replaced it.
3. **The binary is published last.** Companions are placed before it, so a failed required copy leaves the working installation exactly as it was, rather than a new executable beside an old library with the recorded version untouched. The regression test you asked for is the part I could not add honestly: the ordering lives in the `vscode` half, which this suite cannot import — the policy either side of it is tested, and the ordering is stated in a comment naming the state it avoids.

PR #47, opened 2026-09-06
chore(rules): bump the shared-rules pin to c5a6019

oleksandrdubyna88:
Bumps `.claude/rules/shared` to `c5a6019`, which adds **`common/scenario-tests.md`**: every repository owns a scenario harness — in-repo or as a repository of its own pinned by the product — written in the product's language, driving the real product rather than a double, with the flow list derived from the product's own source of truth and catalogued in `research/module_tests.md` (where the harness is, the exact run command, every flow covered / not covered with a reason, what it does not prove, when it runs).

Also in the bump: pointers from `common/testing.md` (plus one DoD line) and `common/knowledge-base.md`, and `todo/PLAN_scenario_harness_check.md` — the plan for `tools/scenario-check.mjs`, which will be adopted with `--warn` first.

Pin-only change: one gitlink line, no repository content touched. Part of the same-task pin cascade required by the conventions README's editing discipline.

Review gate: not applicable — this diff is a submodule pointer, with no code to review.


oleksandrdubyna88:
This PR is correct and cannot go green on its own — and I duplicated its pin bump before spotting it, so here is the state.

The two red checks here (`extension` and `SonarCloud Scan`) are **not** caused by this bump. They are a separate breakage on `main`: `bundledPage.test.ts` pinned its fixture to `2026-09-05T07:43:10Z` while the rounds page opens on *today's* range from the real clock, so the test was green on exactly one day and has been red since yesterday morning. `SonarCloud Scan` runs the extension suite for coverage, which is why it fails too.

That makes a deadlock under `strict: true`: this PR stays red until the test fix lands, and the test fix stays red until this pin lands. Neither is mergeable alone.

#50 breaks it by carrying **both** — the test fix plus this same `c5a6019` bump (cherry-picked, identical one-line change). Once #50 merges, this PR rebases to empty and can be closed.

Also recorded in #50: the cascade was never run anywhere. All six consumers are stale — `creds_for_devs` at `cef6e53`, and `mcp`/`rag_qln`/`benchmark`/`sidecar_rust` two behind at `4dbead4` — so every repository in the family is red on `pin-check` right now, not just this one.

oleksandrdubyna88:
Closing: everything this PR carries is already on `main`, and merging it now would take something away.

#50 merged the identical `c5a6019` bump (cherry-picked from this one's change), so `.claude/rules/shared` is at the remote tip and `pin-check` is green on `main`.

What is left in this branch against the new `main` is not the pin at all — it is a **revert**:

```
src_vs_code/src/test/bundledPage.test.ts | 48 ++------------------------------
1 file changed, 3 insertions(+), 45 deletions(-)
```

That is the branch's older copy of the file, which would undo the fixture fix that made `extension` and `SonarCloud Scan` green and put `main` straight back to red. Nothing to rebase toward — the useful half already landed.

Thanks for spotting the stale pin; the wider finding is recorded in #50: all six consumers were stale, so the other five still need the same bump.

PR #49, opened 2026-09-06
feat(review): the conventions round sees 60 KB of rules, not 40

oleksandrdubyna88:
The conventions round was judging changes against a third of the conventions.

**Measured before changing anything**, on this repository:

| budget | files shown | omitted |
|---|---|---|
| 40 KB (before) | 8 | **19** |
| 60 KB (now) | 12 | 15 |
| the whole set | 26 | 1 (~199 KB) |

At 40 KB the omitted list was the rules findings are actually written against — `testing.md`, `security.md`, `reuse-first.md`, `git-workflow.md`, `utc-timestamps.md`, and all four language doctrines.

This is a budget, not a fix: what does not fit is still **named** in the prompt, so a reviewer cannot read an absence as compliance.

## Verified

- The shared rules do reach the reviewers: the submodule is populated (no missing mounts), and the gate round run minutes before this change quoted `coding-style.md`, `automated-checks.md` and `development-workflow.md` by path and by sentence.
- The test is behavioural, not a restatement of the constant: 55 KB of rule files — six files, the shape of a real family repository — must all fit the default. It goes red at 40 KB with `Expected bundle.Files to contain 6 item(s) ... but found 4`.
- **784 server tests pass.**

## Open question

Even at 60 KB, fifteen rule files never reach a reviewer, and `development-workflow.md` (14 KB) plus `http-contracts.md` (11 KB) crowd out `testing.md` and `security.md` by nothing but collection order. Two candidates, whichever you prefer: raise the budget again, or order the family rules by relevance to the diff (the language doctrine matching the changed file types first).

### DIFF
--- a/src_mcp/runners/Context/RuleFiles.cs
+++ b/src_mcp/runners/Context/RuleFiles.cs
@@ -78,18 +78,21 @@ public static class RuleFiles
         [(".claude/rules", "*.md"), (".cursor/rules", "*.mdc"), (".cursor/rules", "*.md")];
 
     /// <summary>
-    /// Rules are text and a prompt is finite. 60 KB is about a dozen real rule files.
+    /// Rules are text and a prompt is finite. 80 KB is about a dozen and a half real rule files.
     /// </summary>
     /// <remarks>
     /// Raised from 40 KB on 2026-09-06, after measuring what this repository actually shows a
     /// reviewer: at 40 KB it was 8 files and 19 omitted, and the omitted list included `testing.md`,
     /// `security.md`, `reuse-first.md`, `git-workflow.md` and all four language doctrines — the rules
-    /// most findings are written against. At 60 KB it is 12 files and 15 omitted. The whole family
-    /// set is about 199 KB, so this is a budget, not a fix: what the budget cannot fit is NAMED in
-    /// the prompt (the bundle renders an "omitted for length" note) precisely so a reviewer cannot read an
-    /// absence as compliance.
+    /// most findings are written against. Raised to 80 KB the same day, on the operator's call, and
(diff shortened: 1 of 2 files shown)


oleksandrdubyna88:
Raised to **80 KB** on your call, and measured there: **12 files, 77 KB, 16 omitted** — the extra 20 KB buys exactly one more file.

Which makes the ordering the real finding. `development-workflow.md` (14 KB) and `http-contracts.md` (11 KB) are collected first and take a quarter of the budget between them, so `testing.md`, `security.md`, `reuse-first.md` and all four language doctrines are *still* never shown to a reviewer — not because the budget is small but because they are last in line. Ordering the family rules by relevance to the diff would fix what another budget increase cannot; noted in the code and in the PR body rather than done here.

oleksandrdubyna88:
Pushed a commit to this branch and merging it — the quality gate was failing on `new_security_rating 3 > 1`, from two findings in the new `Shuffled`.

**CA1859** is right and costs nothing: the method already builds and returns a `string[]`, so it says so now instead of widening to `IReadOnlyList<string>` for a caller that only foreaches it.

**S2245** ("use a cryptographically strong random number generator") is wrong about this call, and I suppressed it at the site with the reason rather than obeying it. Nothing here guards a secret — the draw decides only *which* rule files a reviewer is shown when the family set does not fit the budget. The `seed` parameter is the tell: it exists so a test can assert an exact order, which a cryptographic generator cannot produce at all. Obeying would have broken the tests and protected nothing. The argument lives at the call site rather than in a settings file, so the next reader meets it where the code is.

`CoaiMcp.Tests`: 784 passed, 1 skipped, 0 failed — 20 of them in `RuleFilesTests`.

PR #52, opened 2026-09-06
chore(rules): bump the conventions pin to c5a6019 — the cascade the rule change left undone

oleksandrdubyna88:
`main` is red, and every PR opened against it inherits the failure. This is the one-line fix.

```
pin-check: STALE .claude/rules/shared
  pinned cef6e53335fd007bc5147698ac6dd9b998fa3503
  remote c5a601907f3313e4b2d9bfd189c6351c6d445943
pin-check: 1 stale pin(s). A rule change and the pin bumps are one task
```

## The wider finding — the cascade was never run

`dew_flow_conventions` moved to `c5a6019` and **no consumer was bumped**. Audited just now:

| repository | pinned at | |
|---|---|---|
| `dew_flow_connect_other_ais` | `cef6e53` | **this PR** |
| `dew_flow_creds_for_devs` | `cef6e53` | stale |
| `dew_flow_mcp` | `4dbead4` | stale, two behind |
| `dew_flow_rag_qln` | `4dbead4` | stale, two behind |
| `dew_flow_benchmark` | `4dbead4` | stale, two behind |
| `dew_flow_sidecar_rust` | `4dbead4` | stale, two behind |

`pin-check` compares against the remote tip, so **every repository in the family is red on this check right now**, not only this one. This PR fixes one of six; the other five need the same bump, and `dew_flow_rag_qln` needs it last because bumping `mcp` and `benchmark` moves the tips it pins in turn. That is a family-wide task, and it is not this PR.

## What adopting `c5a6019` costs

The conventions commit adds `common/scenario-tests.md` (every repository owns a scenario harness and records it in `research/module_tests.md`), the knowledge-base and testing paragraphs pointing at it, and `todo/PLAN_scenario_harness_check.md` — the plan for the check that will eventually enforce it.

**It adds no tool**, so this bump turns nothing else red today. The rule's own *Adoption* section is explicit that a repository without a harness starts by writing the catalogue rather than the tests, on the same reasoning as the `--warn` adoption of the other shared tools: *a check that goes red on the day it lands teaches people to switch it off.*

`research/module_tests.md` is therefore **not** in this PR, deliberately: a pin bump that unblocks the branch and a new required 40-flow inventory are different sizes of change, and mixing them puts an unreviewed document inside a one-line fix for a red `main`.

## Test plan

- [x] `node .claude/rules/shared/tools/pin-check.mjs` → `OK — 1 pin(s) at their remote tips`. Run **after** committing, not after staging: it reads the pin from HEAD, so a staged bump still reports STALE and reads as a second failure.
- [x] `plan-lifecycle.mjs` → OK.
- [x] `gate-snippet-check.mjs` → OK.


oleksandrdubyna88:
Folded into #50. Neither change goes green alone — `SonarCloud Scan` runs the extension suite, so the clock bomb fails it as well as `extension`, and `family checks` needs the pin. Merging them one at a time green-lights nothing, so the pin commit is cherry-picked onto #50 and this PR closes. The family-wide finding (all six repositories stale) is recorded in #50's description.

PR #56, opened 2026-09-06
feat(settings): each side of the machine can keep its own settings

oleksandrdubyna88:
> Stacked on #55 — the diff to read is the last two commits.

One machine, three working environments: a Windows window on one company's subscription and two WSL distros on two others, each needing **its own proxy, its own CLI paths and its own login**.

This was not a bug in the extension — it is how VS Code resolves settings. All nineteen `coai.*` settings are `window`-scoped, and a Remote-WSL window reads those from the **client's** `settings.json` and hands the same values to the remote extension host. Only `machine`-scoped settings can differ per side, and no API writes a remote host's machine settings. So three companies shared one proxy, one set of CLI paths and one vault key.

## What ships

`coai.perSideSettings`, **off by default** — and off is byte-for-byte the old behaviour, no migration, nothing to undo.

On, each side keeps its own values in `globalState` under a key that **reuses the identity the install record already folds**: `remoteName` + distro-or-hostname + storage path. That shape was argued through review once, for the two-sides version record, and the argument holds unchanged here — every WSL distro mounts the same `/home/<user>/.vscode-server/…`, so a storage path alone would merge two companies' settings. It is not written a second time.

- **One accessor for reads, one funnel for writes.** A read that goes around them is a setting that silently stays shared, and the person who set a different proxy would find out when a review ran against the wrong company's server.
- **Turning it on seeds this side from what it reads today**, so nothing changes until something is edited. The seed is idempotent: off and on again keeps what a side had configured.
- **`uiScale` and `helpLanguage` stay shared** deliberately — a text size belongs to the person, not to the company.
- **The panel names the side it is editing.** Somebody with three environments needs to read that off the panel rather than infer it.

## Verified

**475 tests pass.** The ones worth naming, each written for a way this could have been quietly wrong:

- two distros with one storage path are two sides, and the overlay key can never collide with the install record in the one `globalState` they share;
- a **falsy** override is still an override — `??` would have read the shared value for `false`, `0` and `''`, which is every switch somebody turns off on one side only;
- a setting a **later version** adds falls through to the shared value instead of turning itself off on exactly the machines that had customised anything;
- rubbish in the overlay slot (a string, a null, an array from an older build) is no overlay rather than a blank panel;
- a test asserts every setting `settingsFrom` reads is on the overlaid list — one this list forgets stays silently shared.

The manifest description, the `?` tooltip and the help article in **all five languages** say what it does and what it does not touch.

Plan: [todo/PLAN_settings_per_side.md](todo/PLAN_settings_per_side.md). The one open question there — a single shared switch versus a per-side switch — was answered by the operator: one shared switch, every side forks, seeded from what it had.


oleksandrdubyna88:
Pushed a one-line fix to this branch: the rebase onto `main` dropped a test's closing `});` in `install.test.ts`, so the file did not compile (`TS1005: '}' expected`) and both `extension` and `SonarCloud Scan` — which runs the same suite for coverage — failed.

The cause is the rebase, not this branch's work. #45 added `the release smoke fails on the message the server actually prints`, whose body ends in a `for` loop; combining it with this branch's new tests kept the loop's closing brace and lost the test's own terminator. Textually clean, syntactically invalid — the shape a rebase can produce and a merge conflict would not have.

Extension suite after the fix: **475 tests, 474 passed, 0 failed**.

PR #62, opened 2026-09-06
fix(sessions): a session file from an older build no longer kills every code round

oleksandrdubyna88:
**Reported as "coai mcp is dead":** every `review_code` in every repository threw `Value cannot be null. (Parameter 'first')`, while `review_plan`, `status`, `resolve`, `open` and `providers` answered normally.

Reproduced, then diagnosed from the server's own log, which named it exactly:

```
System.ArgumentNullException: Value cannot be null. (Parameter 'first')
   at System.Linq.Enumerable.Union[TSource](...)
   at CoaiMcp.Server.PanelService.<RunStageAsync>d__33.MoveNext()
```

An extension method on a **null receiver** — `session.UsedPrompts`, at the line that records which lenses a round spent.

## The cause is data, not code

`UsedPrompts` arrived on 2026-09-01. A session file written before that has **no such member**, and the source-generated deserializer does not run a property initializer for an absent one — so `= []` was a promise nobody could rely on, and the next round over an old session died. **Nine such files are on this machine.**

New sessions were fine, which is exactly why the plan stage looked healthy: it had just created one. That asymmetry — plan works, code throws — is what made it look like a regression in the code path.

## The fix

Where the collections are **declared**, with the `field` keyword, so no caller has to know which build wrote the file it is reading:

- `UsedPrompts`, `Pending`, `ReviewerStates` — an absent member *and* an explicit `null` both become an empty list. Covering the value matters: one build had already written a null, and without that the bad files stay bad for ever.
- `Rounds` is a **constructor parameter**, so no initializer can catch it. Normalised at `SessionStore.Load`, the single place a session comes off disk. That one would have taken down the panel rather than a round.

## Verified

**RED first — four tests, all failing with the real symptom before the fix:** the exact file shape found on disk, the exact expression that threw, an explicit null, and a session with no `rounds` member. Teeth re-checked afterwards by reverting the normalisation.

**805 tests pass.**

Worth noting for the next incident: the answer came out of the server's log in about a minute — but that log lives *beside the binary* (`globalStorage/.../logs/`), which is not where anyone looks. Moving it next to `coai.db` in the data directory is a small follow-up worth doing.

## Summary by CodeRabbit

- **Исправления**
  - Повышена совместимость с ранее сохранёнными сессиями: отсутствующие или пустые данные корректно загружаются как пустые списки.
  - Устранены ошибки при обработке сессий, в которых некоторые коллекции имеют значение `null`.
  - Загрузка сессий без информации о раундах теперь выполняется корректно.

- **Тесты**
  - Добавлены проверки обратной совместимости для сессий, созданных в старых версиях приложения.

### DIFF
--- a/src_mcp/src/Server/SessionStore.cs
+++ b/src_mcp/src/Server/SessionStore.cs
@@ -58,7 +58,12 @@ public sealed record RoundRecord(
     public int RunnerPid { get; init; }
 
     /// <summary>Per-reviewer progress — the live part of a running round.</summary>
-    public List<ReviewerState> ReviewerStates { get; init; } = [];
+    /// <remarks>Normalised on the way in, for the reason spelled out on PersistedSession.UsedPrompts.</remarks>
+    public List<ReviewerState> ReviewerStates
+    {
+        get => field ??= [];
+        init => field = value ?? [];
+    }
 
     /// <summary>
     /// What this round was ABOUT — the plan's file name or its title.
@@ -93,7 +98,12 @@ public sealed record PersistedSession(SessionState State, List<RoundRecord> Roun
     public DateTime OpenedUtc { get; init; }
 
     /// <summary>The last round's merged findings — what `resolve`'s indices point into.</summary>
-    public List<Finding> Pending { get; init; } = [];
+    /// <remarks>Normalised on the way in, for the reason spelled out on <see cref="UsedPrompts"/>.</remarks>
+    public List<Finding> Pending
+    {
+        get => field ??= [];
+        init => field = value ?? [];
+    }
 
     /// <summary>
     /// The scope: what this change was supposed to achieve, as the plan stage stated it.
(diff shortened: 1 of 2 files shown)


oleksandrdubyna88:
@coderabbitai review

oleksandrdubyna88:
Both applied in `0f5ff3a` — and the first one is the worse half of the defect this PR was opened for. Thank you.

**`SweepOrphanedRounds`.** Right, and I had missed the second reader entirely: it deserialises separately and then calls `session.Rounds.Any(...)`, so an old file would have taken down the sweep that runs at **startup** — every window, not one round. The normalisation is one `private static Normalised(...)` used by `Load` and by the sweep now. Its test drives the real `SweepOrphanedRounds` over the old file shape and goes red without the fix (`NullReferenceException`), which is how I checked the test has teeth rather than assuming it.

**The missing test paths.** Also right, and the reason matters more than the count: absent and explicitly `null` take *different* routes — `get` answers for the absent member, `init` for the null — so covering one proves nothing about the other. And `RoundRecord.ReviewerStates` was not covered at all, though the panel reads it on every repaint, where a null is a blank card rather than a crash: the quiet kind of wrongness that survives for weeks.

Three tests added: the sweep over the old file, an absent `pending`, and `reviewerStates` both absent and null in one session. **808 tests pass.**

PR #71, opened 2026-09-06
feat(mcp): story 1.3 — a Team server is a vendor like any other

oleksandrdubyna88:
## What was asked

Story 1.3 of the Team server plan: the CLIENT half. A company buys one subscription per vendor, installs the vendor CLIs on one Linux VM and runs `coai-server` there; `coai-mcp` on a developer's machine must be able to use that server as an ordinary reviewer vendor. Nothing above the vendor adapter — not the round, the scheduler, the ledger, nor the panel — may learn that a reviewer ran on somebody else's machine.

## What shipped

`remote` is the second runtime that is not a CLI, and it is deliberately built on the shape `local` already proved: the adapter emits a command line that launches this binary in a shim mode (`--ask-remote`), and the shim does the HTTP.

| File | Holds |
|---|---|
| `TeamServerAuth` | one canonical spelling of a server URL, the fingerprint that names its token file, the owner-only read/write of that file |
| `RemoteAsk` | the pure half — request body, poll parse, every exit code and every sentence, so each failure path is tested without a server |
| `RemoteRuntime` | the adapter, and the cancel-an-abandoned-job half |
| `RemoteProbe` | a remote vendor's health over `GET /api/catalog`, with a cache |
| `AskRemote` | the shim itself |

`PanelSettings` needed **no change at all** — `RuntimeOf` already derives from `ReviewerRuntimeSelector.RuntimeNames`, which is exactly why that set exists after two hand-written copies of it drifted.

## What the tests showed

Observed, not intended:

- **921 MCP tests pass**, 0 failed, 1 skipped (the pre-existing elevation-gated symlink test). 816 before this branch.
- **165 server tests pass**, unchanged — `VendorProbe` is shared with that binary.
- `plan-lifecycle`, `pin-check` and `gate-snippet-check` all clean.

Four guards were proved to have teeth by breaking the code they cover and watching the real symptom:

| Broken | Red said |
|---|---|
| the `remote` arm moved after the base-URL arm | `Expected NameOf(vendor) to be "remote" ... but "codex"`, and its auth became a vault key |
| the trailing-slash trim removed | `…com//api/catalog`, and two fingerprints for one server |
| the abandon call removed from the timeout path | `Expected spy.Abandoned to be 1 ... but found 0` |
| (fixture) a missing brace | seven probe tests, naming the unparsed catalog |

`RemoteShimScenarioTests` runs the **real** `coai-mcp --ask-remote` binary in its own process against a real `HttpListener` on a real socket — the answer file, the usage line, the exit codes, the URL that goes on the wire, and both cancellation paths. The shim had never been launched by any test; this repository has twice shipped a break of exactly that shape with every in-process test green.

## The review gate

`review_plan` reached `proceed`; `review_code` ran over this diff before this pull request was opened, nine reviewers across three vendors, 27 findings — **20 accepted and fixed, 7 rejected with reasons**, all recorded on the round.

The most valuable finding was that **the story's own requirement 7 was not met**: the machinery for cancelling a review whose shim had been killed existed, and nothing called it. The job file path never left `Build`. A killed reviewer ran to completion on the company's subscription for an answer nobody would collect — the exact outcome the mechanism was written to prevent. Fixed with an `AbandonAsync` seam on `IReviewerRuntime` (default `Task.CompletedTask`, so no CLI adapter changed) invoked from `ReviewerExecutor.LaunchAsync` on both kill paths. Three further defects in the same mechanism would each have made it look wired and work on nothing: the `DELETE` omitted `X-Coai-Contract` (the server answers 426 without it), the claim was deleted in a `finally` so a brief network drop lost the job id for ever, and the claim could not authenticate itself.

### Not acted on, and why

- **"Unreadable usage is reported as zero."** This is the repository's existing convention, not something this change introduced — `LocalRuntime.ReadUsage` returns the identical `Usage(0,0,null)`. Distinguishing an unknown measurement from a real zero is a change to the `Usage` type and every consumer of it, which is a plan rather than a hunk in a story about a new runtime; a one-adapter fix would produce two conventions and no benefit.
- **"`status: cancelled` falls through to Queued and hangs."** The premise is wrong on this server: `JobStatus` is `queued|running|done|failed` and a cancelled job arrives as `failed` with failure `cancelled`, already handled. The robustness point behind it was raised separately, accepted, and fixed with an explicit `Unknown` state.
- **"Null `RemotePoll` crashes or loops."** Already handled — `PollOnceAsync` returns an exit on null and `PollAsync` reads it through a null-check pattern.
- **"Derive the runtime list from `RuntimeNames` instead of `InlineData`."** The derived guard the rule asks for already exists two tests below the cited line, and was added for that rule when `local` drifted.
- Three more (a test file's class name, a message said to lack context it already contains, and two duplicates) — reasons recorded on the round.

## Scope note

The extension's declaration points (`models.ts` RUNTIMES, `vendors.ts`, the `package.json` enum) are **deferred to story 3.1** by an accepted finding from the plan round: a `remote` option a user cannot configure — no server URL field, no sign-in, no token — would always exit `77 not signed in`.


oleksandrdubyna88:
Pushed `8bb1aa9`: a self-audit against `.claude/rules/shared/csharp/doctrine.md` §6 (cyclomatic complexity ≤ 4) found four of this branch's methods well past it, and decomposing them is also what answers SonarCloud's one failing condition.

**The gate said:** new-code coverage **79.3%** against its 80% threshold. Everything else passed — reliability, security and maintainability all rating 1, duplication 0.7%, security hotspots 100% reviewed.

**Why that number was misleading, and why I did not waive it.** `AskRemote.cs` measured worst at 41 of 141 lines uncovered, because its real exercise happens in `RemoteShimScenarioTests` — which launches the shim as a *separate process*, where the coverage tool cannot see it. The lines were tested; they were not *measurable*. Waiving the threshold would have been defensible and wrong: the honest reading is that the polling loop's branches were reachable only through a subprocess, and a subprocess cannot be steered into "three dropped polls in a row" on demand.

So the loop's decision moved into `Decide` — a pure switch expression over one poll's outcome — and its bookkeeping into a `PollTracker` record. `RemotePollDecisionTests` is 27 tests that could not have been written before that change.

**Observed:** 951 MCP tests pass, 0 failed, 1 skipped (up from 921); 165 server tests still pass. The 88 pre-existing remote tests were green both before and after the decomposition, which is what says the behaviour is unchanged.

**Not done, deliberately:** `VendorProbe.RunAsync` breaches the same complexity rule at ~8 branches and predates this branch. This change only added one arm to it; rewriting it here would turn a story into an unrelated diff. Naming it rather than fixing it silently or imitating it — happy to do it as its own change if you'd prefer.

PR #82, opened 2026-09-07
fix(panel): the Team-server sign-in sends its token where the server reads it

oleksandrdubyna88:
## The defect

**Signing in to a Team server could not work in 0.31.1, for anybody.**

`createSession` posted the Microsoft token as `{ token }` in the request **body**. `POST /api/session` is authorised like every other route on that server: `Auth.Bearer` reads `Request.Headers.Authorization`, and nothing on that server reads a token out of a body. So `RequireCaller` found no identity and answered **401 — before any JWT scheme ran**.

## Found on the deployment, not inferred

The operator's own attempt, in the server journal:

```
POST /api/session - application/json 2114  ->  401 0 null 0.6838ms
```

2114 bytes is the real Microsoft token, in the body — and there is **no `JwtBearerHandler` line against that request at all**. The absence is the diagnostic: a token that is examined and rejected always logs an `IDX…` failure and costs milliseconds, not fractions of one.

Both directions were then reproduced against `coai.remsoft.dev`:

| request | result | handler ran? |
|---|---|---|
| POST `/api/session`, token in the body, no header | `401 0 null 0.3504ms` | **no** |
| POST `/api/session`, `Authorization: Bearer …`, no body | `401 0 null 2.5957ms`, `IDX10223: Lifetime validation failed` | **yes** |

The header path is the live one; the body path never reaches authentication.

## Why it was unreadable from both ends

The 401 challenge carries no body, and `said()` falls back to `the server answered ${status}` exactly when the response is empty. So the panel could only report *"RemSoftDev: the server answered 401"* — a sentence that names neither the token nor the reason. On the server side the refusal is silent by construction, because nothing was ever authenticated to fail.

Ruled out along the way, each with evidence rather than assumption: the Entra pre-authorisation (`aebc6443-…` is listed against `coai.access`), the audience shape (`Auth__Microsoft__Audiences` carries **both** `<client-id>` and `api://<client-id>`), the tenant, and the account (a wrong domain answers **403**, not 401 — `Auth.RequireCaller` separates the two).

## The fix

One line on the client side of the contract: the token goes in the header, where every other call on this API already puts it. The server is unchanged — its model is one authentication path, and a second one that reads bodies is the thing this codebase deliberately does not have.

## Tests

**606, 0 failing** (605 before; one added).

Red before green, per the repository rule. The new test failed against the unfixed code with `+ undefined  - 'Bearer idp-token'` — the defect itself, not a setup error.

The old assertion is the reason this shipped: `SessionTests` posts a header and `content: null`, while `teamServerApi.test.ts` asserted the token was **in the body**. Two green suites, contradictory expectations about one seam, and nothing crossing between them. That assertion is now the header, and the body is asserted **absent** — a bearer token in a request body is a bearer token in a log.

## Docs

`research/module_team_server.md` gains the rule (*every route reads its credential from `Authorization` and nowhere else*) and a note on how the failure presents, because a bodyless 401 with no handler line is hard to read from either side.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed Team server sign-in failures by sending identity-provider tokens through the `Authorization` header.
  - Sign-in requests no longer include tokens in the request body, preventing premature `401 Unauthorized` responses.

- **Documentation**
  - Documented the authentication requirements and the sign-in compatibility issue.

- **Chores**
  - Updated the VS Code extension version to 0.31.2.

### DIFF
--- a/research/module_team_server.md
+++ b/research/module_team_server.md
@@ -57,8 +57,11 @@ sequenceDiagram
 | `DELETE /api/reviews/{id}` | owner | `204` |
 | `GET /api/usage?window=&scope=` | any / **admins for `company`** | per-vendor totals for the caller, or for everyone plus per person · `400` unknown window · `403` company as a non-admin |
 
-**Every route reads its credential from `Authorization: Bearer …` and from nowhere else** —
-`Auth.Bearer` looks at that header, and no handler here reads a token out of a request body.
+**Every route that has a caller reads its credential from `Authorization: Bearer …` and from
+nowhere else** — `Auth.Bearer` looks at that header, and no handler here reads a token out of a
+request body. The two anonymous routes above (`/api/health`, `/api/client-config`) have no
+credential to read at all, deliberately: a client that has not signed in yet must be able to ask
+this server what it wants.
 `POST /api/session` is the one people expect to be different, because it is where a session begins;
 it is not.
 
--- a/research/module_team_server.md
+++ b/research/module_team_server.md
@@ -57,6 +57,21 @@ sequenceDiagram
 | `DELETE /api/reviews/{id}` | owner | `204` |
 | `GET /api/usage?window=&scope=` | any / **admins for `company`** | per-vendor totals for the caller, or for everyone plus per person · `400` unknown window · `403` company as a non-admin |
 
+**Every route reads its credential from `Authorization: Bearer …` and from nowhere else** —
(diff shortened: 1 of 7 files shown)


oleksandrdubyna88:
Applied in `763f18d`, with the wording widened a little past the suggestion: the sentence now says *every route that HAS a caller*, and names the two anonymous routes and why they are anonymous — a client that has not signed in yet must be able to ask this server what it wants. The contradiction with the table was real.

PR #83, opened 2026-09-07
test(contract): the real client against a real server, so two suites cannot disagree

oleksandrdubyna88:
## Why

0.31.1 shipped a Team-server sign-in **nobody could use**. The extension posted the identity token in the request **body**; the server authorises `POST /api/session` from the `Authorization` header alone. Both suites were green about their own half — `SessionTests.cs` posts a header and a null body, `teamServerApi.test.ts` asserted the token was in the body — and neither can observe the other.

The instance was fixed in 0.31.2. This closes the **class**.

## What runs

`src_vs_code/src/contract/teamServerSession.contract.ts` drives the extension's **own** `createSession` and `ask` against a **real** server:

| assertion | why it is there |
|---|---|
| a session is minted — `201` | the regression itself; `401` here whenever the halves disagree |
| the issued token works on a later call | proves the whole handshake, not just the first hop |
| an outside domain is refused **403, not 401** | the teeth: proves the server is really authorising, and pins the distinction that made the original defect readable |

**No Microsoft.** The server's symmetric `Local` scheme exists for exactly this (`Auth:Local:SigningKey`, issuer `coai-local`, HMAC-SHA256 over ≥32 bytes). Minting a token is fifteen lines of `node:crypto`, no dependency — which is what makes this cheap enough for every PR.

## Where the server comes from

`scripts/run-contract.mjs`:

- **`COAI_CONTRACT_URL` set** → uses the server you already have. That is the `deploy/docker-compose.yml` story, unchanged: `docker compose up` then `npm run test:contract`.
- **unset** → starts the server this repo just built, on a free loopback port, with a throwaway data directory and a generated key; stops it afterwards.

Under `npm run test:contract`, deliberately **not** `npm test` — that suite stays hermetic and needs nothing running.

CI runs it in `build · test · family checks`, which already has the .NET SDK and has just compiled the server. **The container was rejected for CI on cost, not principle**: `src_server/Dockerfile` publishes Native AOT (clang, a full link) and would add minutes to every PR, and there is no image to pull instead — no `server-v*` tag exists. It stays the better local story, which is why the address is an input.

## Proof it has teeth

Measured, not assumed — revert the 0.31.2 fix and run it:

```
✖ the real client mints a session against the real server
  AssertionError: the server refused the sign-in: status 401: the server answered 401
✖ the session the server issued is accepted on a later call
✖ an account outside the company domain is refused 403, not 401
  tests 3 · pass 0 · fail 3
```

The first failure is **the operator's own sentence**, reproduced from the live defect. Restore the fix: `3 pass, 0 fail`.

The third going red is its own finding: with the token in the body the server cannot even tell the domain is wrong, because nothing was authenticated at all.

## Also here

**`chore(rules): bump the conventions pin to c4e521f`** — a separate commit, and not optional: `pin-check` was red on *every* branch in this repo because the conventions submodule moved and the cascade did not reach this consumer. The whole diff between the two pins is one Dependabot commit against the conventions repo's own workflow — **no rule text changes** — which is why it is a one-line bump rather than a cascade. (The existing `chore/conventions-pin` branch does not help: it targets `c5a6019`, which the remote has moved past, on commits that are not on main.)

## Observed

`tsc` clean · **606** unit tests, 0 failing (unchanged — contract files compile to `out/contract`, `run-tests.mjs` reads `out/test`) · **3** contract tests, 0 failing · `npm run package` produces the .vsix.

Two things shipped differently: the assembly is `coai-server.dll`, not `CoaiServer.dll` (the project sets `AssemblyName` — the script's error names both paths it looked in, which is why that cost one run rather than a session), and logs need no switch because `CoaiLogPath.RootFor` puts them under the data directory.

## Docs

The plan is promoted to `research/PLAN_contract_across_the_seam.md` with the layer decision recorded, and its two unbuilt DoD items are extracted into `todo/PLAN_refusals_that_explain_themselves.md`: a refusal that says **why**, and the one-credential-path rule enforced by a test rather than a paragraph.

## Summary by CodeRabbit

- **Tests**
  - Added end-to-end validation of the sign-in and session flow using the real client and server.
  - Confirmed authenticated follow-up requests succeed and users outside the permitted company domain are rejected.
  - Integrated contract testing into the continuous integration workflow to detect client/server compatibility issues earlier.

- **Documentation**
  - Added implementation and planning documentation for contract testing and clearer authentication refusal behavior.
  - Updated research and task indexes to reflect completed and upcoming work.


oleksandrdubyna88:
All seven review threads are addressed in `a56f0da` and `4b8a143`. Two of them were things the multi-vendor gate missed, and one was the best finding on this branch from any source.

| Thread | What happened |
|---|---|
| `ci.yml` — pin `actions/setup-node` | **Pinned** to `8207627…` with a `# v7` comment, matching the release workflow's house style. This is also what took SonarCloud's new-code security rating to C, so it fixed the quality gate too. The two older unpinned `@v7` lines are named in a comment as a separate change rather than swept into this one. |
| `run-contract.mjs` — clean up failed local-server starts | **Fixed**, and it was worse than the thread said: the SIGKILL escalation was `unref`'d with `process.exit` on the next line, so the timer meant to catch a server ignoring SIGTERM could never fire. Teardown is now one idempotent handler reached from `exit`, the signals, `uncaughtException` and every `fail()`, taking the process **tree** — a process group on POSIX, `taskkill /T` on Windows. |
| `run-contract.mjs` — handle test-child start errors | **Fixed.** Nobody else caught this: `spawn` failure emits `error`, not `exit`, so the promise would never have settled and the command would have looked frozen. There is now an `error` listener and a whole-run deadline, because a hang while resolving an import never reaches `--test-timeout`. |
| `run-contract.mjs` — health request timeout | **Fixed** with `AbortSignal.timeout`; a server that accepts and then says nothing could otherwise park the readiness loop past its own deadline for ever. Readiness came down from 60s to 30s and prints what it is waiting for. |
| `run-contract.mjs` — the missing-key message | **Rewritten** to say *why* the script will not generate a key for a server it did not start: one it invented would not match, and every assertion would fail as a 401 that looks like a contract break. |
| Place the contract test under `src/test` | **Moved.** It stays out of `npm test` because that runner lists `*.test.js` and this compiles to `*.contract.js` — the separation is the suffix, not the folder. |
| Expose and assert the session response status | **Done, and thank you** — this is the best finding here. The `ok` arm of `ServerResult` discarded `response.status`, so no caller could tell `201 Created` from any other 2xx, while the server's own test asserts exactly `201`. The two halves of this contract could have agreed about the body and disagreed about the code: the same class of gap this branch exists to close, one field smaller. `status` is on both arms now and the test asserts `201`. |

Re-proved after all of it rather than assumed: reverting the 0.31.2 fix turns all three contract tests red, restoring it turns them green. `tsc` clean · 606 unit tests, 0 failing · 3 contract tests, 0 failing · both workflows parse · `plan-lifecycle`, `pin-check`, `gate-snippet-check` clean · no orphaned server process after a run.

PR #100, opened 2026-09-08
feat(runners): a round's Team-server reviewers are not submitted in list order

oleksandrdubyna88:
Two things the operator asked for: the load skew, and the "haiku is slow" question.

## 1. The queue is not served in the order everybody typed

A Team server holds **one company account per vendor** (`slotConcurrency: 1`), and every client submits a round's reviewers in the **same order** — because every vendor list is built the same way: the catalog's order, appended as vendors are added, preserved by the settings file, walked by the fan-out.

With one person that is invisible. With ten submitting inside a minute, the vendor sitting first in everyone's list takes ten reviews into a queue of depth one while the third vendor's account is idle, and each of those ten waits behind nine strangers for an account that could have answered at once. **The load was skewed by a list order nobody chose.**

`SubmissionOrder.For(work, roll)` — pure, randomness injected — permutes the **remote** entries among the positions remote entries already hold. `BoundedScheduler` dispatches in that order and puts results back into the caller's, because the two requirements pull against each other: the load fix wants the start order to vary, and a person watching must not see their reviewers move between runs.

**`JobStore.TryClaim` is deliberately untouched.** Fairness there is what makes a queue position mean anything; randomising the *claim* would turn a long wait into an unpredictable one. `TheOldestQueuedJobGoesFirst` already guarded it and now says why it must not be relaxed.

Red first, with the real symptom: 40 consecutive rounds, three Team-server reviewers, `found 1: {"s-a"}`. The distribution test asserts **flatness within a tenth of the fair share over 3000 rounds** — "not always the same vendor" would pass a shuffle sending 80% of first submissions to one account.

## 2. "Haiku is slow" — measured, and refuted

`research/RESULTS_haiku_is_not_slow_it_is_verbose.md`, from 57 real reviews on the deployed server:

| vendor | n | median out | median seconds | **ms per output token** |
|---|---|---|---|---|
| claude · haiku | 15 | **11 950** | 135.1 | **11.67** |
| codex | 24 | 1 948 | 46.7 | 26.16 |
| antigravity | 18 | 1 102 | 12.4 | 11.44 |

Haiku generates at antigravity's rate and at **less than half codex's cost per token**. It is not slow — it writes **6–11× more** from the same prompt.

And the `in` column that suggested otherwise is three different quantities: claude counts `input + cacheCreation + cacheRead` summed across models (so a cached prefix is re-counted every turn), antigravity omits cache entirely, codex reports a cumulative maximum. An identical trivial review measured **29 513** input on claude against **12 593** on codex — all harness, cached and read back.

Deliberately **not** "fixed" here: `maxItems` on the findings array would cut real findings (nothing downstream discards them — `maxPerProvider` is a concurrency cap), and normalising the usage columns is its own change, `todo/PLAN_usage_that_compares.md`.

## Gate

Plan round: 8 gating, 4 accepted. Code round: **`proceed`, all 9 reviewers answered**, 8 gating, 4 accepted. Both rounds' rejections carry reasons recorded through `resolve` — including two "Blocking" findings that were factually wrong about the code (`_roll` is a field at `BoundedScheduler.cs:76`; a shared `Random.Shared` does not produce identical permutations, which the paired test measures rather than assumes).

MCP **1053** passed · server **181** · extension **647** · plan-lifecycle clean.


oleksandrdubyna88:
Closing: this feature was built twice in parallel, and the other implementation is better.

While this branch was in review, another session shipped the same capability on `main` — `SeededShuffle` in the core, applied in `PanelService.BuildWork`, with `research/PLAN_vendor_submission_order.md` promoted. Reuse-first is explicit that a second implementation of a capability is a defect from the moment it compiles, so this one goes rather than being merged beside it.

Theirs is the better design on one clear axis: the shuffle is seeded from the session id and the round number, so two sessions differ while ONE session replays to the same order. This branch used `Random.Shared`, which makes a round unreproducible to buy nothing.

What survives is the half that was not duplicated — the haiku measurement and the usage-accounting plan — reopened as #101.

PR #103, opened 2026-09-08
fix(bench): a Team-server vendor keeps the name its server knows

oleksandrdubyna88:
## Found by the load campaign, in its first twelve rounds

Every one returned `call_human` in about seven seconds with zero tokens, and the log carried one sentence thirty-six times:

```
'remsoftdev-claude' is not a vendor here
```

## The defect

The bench rebuilds the vendor list for the server it starts, and `VendorConfig` had no `remoteVendor` — so the server was handed the **row id**. A row is named `<server>-<vendor>` precisely so two Team servers each offering `codex` do not collide on one id, and no server has ever heard of `remsoftdev-claude`. Every remote reviewer was refused with a 400 before it began.

**The bench could not measure a Team server at all**, and nothing said so: twelve rounds, three vendors each, a full report with tables — and not one reviewer had run.

## The shape to remember

This is the same field, the same failure and the same discovery route as [the extension's version of it](https://github.com/oleksandrdubyna88/dew_flow_connect_other_ais/blob/main/research/architecture.md), which shipped for the whole life of the `remote` runtime.

And it is exactly what `Vendors.cs`'s own docstring warns about — *"a bench that rebuilds vendors from what it happened to read measures a machine nobody has"*. The field was added to the panel **after** this reader was written, and nothing carried it across. That is the shape to look for, not the field.

## Tests

- the name is read from the settings file
- it survives being serialised back for the server — reading it is only half; the round trip is where it was lost
- a vendor that is not remote still says nothing about it, written only when there IS one, exactly as the panel writes it

## Test plan

- [x] 112 bench tests, 0 failing
- [x] the campaign re-run reaches real reviewers instead of a 400

## Summary by CodeRabbit

- **New Features**
  - Remote vendor names are now preserved when vendor settings are read and saved.
  - Local vendors continue to omit empty remote vendor information.

- **Documentation**
  - Added load-test results covering Team server capacity limits, queue behavior, timeouts, throughput, and recommended next steps.
  - Updated the research documentation index to include the new load-test report.

- **Tests**
  - Added coverage for preserving remote vendor names and omitting them for local vendors.


oleksandrdubyna88:
Five threads read; three were right and are taken in `ad0d58b`.

**`AsSetting` gated on the length alone.** `Vendors.Read` preserves whatever a settings file carries, so a row that used to be remote and is now local arrives still holding the name — and the field would have been written back out for it. That is the same silent field mismatch this pull request exists to end, pointing the other way. Gated on the remote runtime now, with the runtime word as a named constant rather than a literal, and the regression test uses a LOCAL vendor holding `RemoteVendor = "stale"`. RED confirmed by removing the gate: the new test fails and the old one does not, which is what made it worth adding.

**The load record contradicted its own table.** It said six rounds pinned at the bound because "those reviewers ran out of time"; the table below it counts 7 refusals with 429, 3 queue deadlines and 2 reviewer timeouts. The sentence points at the table now instead of naming one cause. It also says which runs it covers: ten of twelve, the ones that actually ran in parallel — the command was 4 cases × `--repeat 3` at `--parallel 10`, and the trailing two never faced simultaneous load, so they are in no table here.

The two `Vendors.cs` threads were the same finding twice and are answered by the one gate.

113 bench tests, 0 failures. `plan-lifecycle: clean`.

PR #150, opened 2026-09-09
feat(log): the page asks for a page, and lets SQL do the counting

oleksandrdubyna88:
## The instruction

From the operator, verbatim: *«возвр все из БД не нужно. у нас есть пагинаций. 200 на стр
достаточно.»* and *«суммы - скл счиатть (сколько всего и тд.)»* — do not return everything from the
database, there is pagination, 200 per page is enough; and count the totals in SQL.

## The measurement that decided the shape

`coai-mcp --log --limit 300`, against the real database:

| part | size | share |
|---|---|---|
| whole payload | **3.83 MB** | |
| the 236 round rows, findings stripped | **0.05 MB** | 1.3 % |
| the 3 484 findings carried inside them | **3.78 MB** | **98.7 %** |

The payload was never the rounds. It was the findings, shipped for every round whether or not
anybody opened one — and the page opens them one row at a time, on a click.

## Server

- `--log --paged` answers a page of rounds with **no findings inside them**. Default 200, max 1000;
  a limit is clamped rather than refused and an unreadable cursor asks for the first page.
- **Keyed, never `OFFSET`.** Rounds are inserted at the top of `started_utc DESC`, so an offset
  shifts every later page when a round finishes mid-read. The cursor is the pair `started_utc|id`,
  because two rounds can start in the same second.
- Each listed round carries `FoundCount` — without it `statusOf` cannot tell a gate still open from
  a round that raised nothing, and every clean round would read as awaiting a resolve nobody owes.
- `--findings --session --stage --number` answers one round's findings. **Three exit codes, because
  there are three answers:** `0` and a list, `69` for a round the database has never heard of, `74`
  for a database that could not be read.
- `LoggedTotals`: two statements, one scan each.
- **Without `--paged` the binary answers exactly what it answered yesterday**, findings inline, from
  its own single grouped query.

## Extension

- The table pages at **200** with Newer/Older and a line saying which rows those are; under it, what
  SQL counted over the whole table.
- **An opened row is honest about all five states**: asking, failed with a retry, none, not kept,
  and the list. Four of them were one blank, and a blank reads as "this round was clean". This
  closes the one item `PLAN_findings_in_the_log.md` never built.
- Filters, search and sort are page-local and return to the first page.
- `--paged` is the compatibility hinge; all four extension/binary pairings are tested.

## What the code round changed

Twelve reviewers, thirty-five findings, thirteen taken. Four would have shipped a defect: `69`
shared between "no such round" and "unreadable database"; a malformed answer parsed into an empty
list, which is a claim that a round was clean; a new extension asking `--findings` of an old binary
and turning an honestly clean round into a failure; and the compatibility path doing one findings
query per round. Plus the shared mutable global — the round's identity now travels with the request,
and the read is wrapped so a rejection cannot leave a row reading "Reading…" for ever.

## Deviations the code found and the plan did not know

**The page's rows are not database rows.** `rowsFrom` builds every row from the session files and the
database only enriches them, so paging the database would not page the page — it would take the ✓/✗
badge away from older rows. Hence two axes: the list is what enriches (1000 rounds, ~220 bytes each),
the page is what a person reads (200). `FoundCount` had to be invented for the same reason.

## Tests

RED first and recorded: 9 of 10 server tests failed before the implementation, each naming its own
defect — `found 250` where 200 was asked for, `{1,2,3,1,2,3,…}` for a cursor that could not pass a
tie, `Known to be True, but found False`. The extension's fallback test was proved to have teeth by
breaking the branch: `actual: 1, expected: 2`.

`LogCliScenarioTests` runs the real binary for `--log --paged`, `--log` unpaged, `--findings`, a
round the database does not hold (69) and an unknown flag (64) — the exit codes are the whole
interface and could not be checked any other way.

**1038 extension tests, 1128 server tests, 0 failures.** `plan-lifecycle: clean`.

## Also in this branch, deliberately

The first commit promotes two plans that had shipped through a different door than they were written
for (`PLAN_findings_in_the_log.md` — its gate showed `resolve` empties `pending`; `PLAN_local_db_reader.md`
— its gate showed a `sql.js` snapshot cannot see `coai.db-wal`) and extracts their unfinished halves.
The planning-docs rule requires promotion in the same task as learning a status line is wrong, and
this work is what learned it: one of the two tails is the five-state row built here.

## Open tail

A row whose round is outside the 1000-round window loses its accepted/rejected badge until somebody
opens it. Closing it means asking the database for the decision counts of exactly the rows a page
holds — a per-page read this change deliberately did not add.

## Summary by CodeRabbit

- **New Features**
  - The rounds log now supports 200-item pages with Older/Newer navigation and whole-log totals.
  - Findings load when a round is opened, with clear loading, empty, unavailable, and retryable failure states.
  - Added paginated log and single-round findings CLI modes, including cursor navigation and bounded limits.
  - Older extension and server versions continue to work through compatibility fallbacks.

- **Documentation**
  - Updated release notes and research documentation to reflect the delivered log, pagination, and findings functionality.
  - Added planning documentation for future findings search.


oleksandrdubyna88:
Thirteen threads read; ten taken in `8fea9c9`. **Two of them would have shipped the feature broken**, and both are worth naming.

**Every row opened saying the database had no record of it.** The page is painted before the database is read — reading spawns a process, and nobody should wait on one to see their log — and an unread log was indistinguishable from a read empty one, because `0 >= 0` is true. So the first paint called every round *absent*, and the tick's merge then held that state against the fresher rows that knew better. `DbLog.read` now says whether the database was asked at all, and the merge carries only what a READ produced — loaded, asking, failed — letting absent and unasked be recomputed from the fresher log. RED confirmed by reverting the fix.

**The totals were stored and pushed nowhere.** `update` recorded them in `latest`; no region carried them and the page's handler had no case for them, so the SQL-counted line under the table would have been empty for ever. It is a region now, delivered by the same push machinery as the rows.

The other eight: the legacy `--log` default fell from 300 to 200, which quietly takes a hundred rounds off the only list an old extension can show (`LegacyLimit`, with a CLI scenario over 201 rounds); a paged read of an unreadable database exits 74 while the legacy shape keeps exit 0; `Cursor` validated only the number after the separator, so `0000|1` compared `0000` against `started_utc` and answered an EMPTY page instead of the first one; a loaded row whose count disagrees with what arrived drew "Reading…" for ever because `ask` refuses a loaded row, and now says what happened with a retry; the totals test used one round and a limit of 1, which passes just as well if aggregation were page-scoped, so it uses two with the page holding one; the CLI scenario read stdout to the end before waiting, which deadlocks when stderr fills, and left a timed-out process alive — both pipes drained concurrently, child killed on the deadline; and four documents contradicted the code they ship with.

Three were declined and the reasons are in the resolved threads.

1041 extension tests, 1130 server tests, 0 failures. `plan-lifecycle: clean`.

PR #169, opened 2026-09-09
feat(chat): a model's answer, rendered as a document

oleksandrdubyna88:
The **renderer half** of [PLAN_an_answer_reads_like_a_document.md](todo/PLAN_an_answer_reads_like_a_document.md) — `renderAnswer` and its hostile-input file. Nothing is wired to it yet.

## The problem

An answer arrives as markdown and reached the chat tab as the raw characters under `pre-wrap`: a numbered list as `1.` down the left margin, a code block as backticks, a heading as hashes.

## The decision went against the plan, and the gate is why

The plan recommended **`marked`** with a custom renderer, as this extension's *first runtime dependency*. I proposed the opposite at the plan round and asked to be corrected if wrong. **I was, in part.** gemini raised a Blocking finding that escape-first is not the invariant I claimed: a rule running over pre-escaped text is reading *entities*, not the characters the model wrote — a `>` blockquote marker is already `&gt;`, a code span carrying `&` is already `&amp;`. That is correct, and it was the technique I had proposed.

Its own fix offered two branches and **I took the second: tokenise the raw text and escape at emission, per context.** Text through `escapeHtml`; an `href` through a **scheme allow-list first**, then `escapeHtml — so a target that is not `http(s)` never becomes an attribute at all. That keeps `javascript:`, `data:` and `vbscript:` out by construction rather than by a filter that enumerates them.

I declined the dependency half, with reasons recorded: zero runtime dependencies in a VSIX installed on developers' machines; safety as a property one function proves rather than a configuration that must stay right across upgrades; a bounded construct set with a literal-text fallback, which is exactly the case a full CommonMark parser is *not* needed for. **The extension still has zero runtime dependencies.**

## The hostile file earned itself on the first run

The first version of `inline` tokenised correctly and then emitted the surrounding prose **raw** — it had no escape of the remaining text at all — and `<script>alert(1)</script>` reached the page. Thirty-four hostile cases caught it immediately. That is the defect the gate's finding predicted, arriving by a route neither of us named.

**The assertions were wrong too, and are now right.** They searched the output for scary substrings (`javascript:`, ` onerror=`, `<script`), which *fails on correctly-escaped answers*, because escaped text legitimately contains those words when the model is writing about them. They now **parse** the output: after escaping, a literal `<` can only be a tag this file emitted, so every tag name and every attribute is checked against an allow-list, with `href` restricted to two schemes. Stronger, and free of false positives.

## What the code round then found — one that no escaping test could

**A code span inside a link label leaked four raw control characters and lost the span.** Code spans are held aside first, so ``[`renderAnswer`](url)`` held the span, then held the *link* with that placeholder inside it — and the restore was a single pass over the outer string, which never looks inside a construct it has just restored. It now repeats until no sentinel remains, bounded by the number of held constructs.

Also: an ordered list swallowed an unordered one following it at the same indent; the closing-fence regex was recompiled per line of a code block; nested list items cloned the parent's child array per child (O(N²) on a long checklist); and `fileHref` was dead code prescribing an `href` form the contract says not to use.

## What is supported, and what happens to everything else

`SUPPORTED` is an exported constant, so the test walks it and the list cannot drift from the code. Anything outside it is **not parsed and not dropped** — a table renders as its pipes, a blockquote as its `>`, a task list as `[ ]`. Ugly and honest, and a far smaller failure than a half-parsed table: the cells are still readable.

**A file reference carries `data-file` and `data-line` and no `href` at all.** A pure function cannot know whether a path exists, and a reviewer was right that deciding "resolvable → link, otherwise plain text" inside it would be a promise it cannot keep — the page decides on a click. **Nothing autolinks a bare URL**; this family shipped that defect once.

## The gate

- Plan round: `good_enough`, 2 of 3 reviewers (codex rate-limited), 8 gating — 8 accepted, 1 rejected.
- Code round: **`proceed`**, 6 of 8 reviewers, 9 gating — 6 accepted, 9 rejected with reasons.

## Tests

RED first, and honestly red: 13 of 57 failing at first, including the real escaping defect; then 3 more for the code round's findings, with the sentinel leak reproduced exactly. GREEN: 61 of 61 in the renderer's own file, `npm test` **1206 tests, 1205 pass, 0 fail, 1 skipped**. `plan-lifecycle.mjs` and `pin-check.mjs` clean. `research/module_tests.md` gains the flow row.

## Not in scope

`chatMessagesHtml`, the styles, the sides and colours, the end rule, the copy control and the `openExternal`/`openFile` handlers — all in `chatPage.ts` and `chatCommand.ts`, which the chat-page lane owns. Nothing calls `renderAnswer` yet, so nothing a person can see changed: no setting, no command, no manifest entry, no help article and no CHANGELOG entry.


oleksandrdubyna88:
Closing as duplicate. `PLAN_an_answer_reads_like_a_document` shipped whole in #168, which is on `main` with `renderAnswer.ts`, the hostile-input file, the link handling, the sides, the end rule and the copy control. This branch built only the renderer half, and building it separately was a mistake on my part about lane ownership rather than a disagreement about the work.

**#168's implementation is the better one and should stand.** It takes `marked` as a runtime dependency and lets a real CommonMark parser do the parsing, which handles nested lists and tables correctly. Mine deliberately avoided the dependency and paid for it by rendering anything outside a bounded construct set as literal text — a table came out as its pipes. That trade is defensible in isolation, and it is plainly worse than parsing the table.

The two review gates reached different answers from the same plan, which is worth recording: the plan recommended `marked`, #168 took it, and my round's security finding pushed me to tokenise-and-escape-per-context without a dependency. Both are safe; only one of them renders a table.

**Two hostile-input cases here are not in #168's file**, and they are the only thing in this branch worth carrying across:

```
{ name: 'a form and an input', markdown: '<form action="https://evil.example"><input name="p"></form>' },
{ name: 'a comment that tries to swallow the page', markdown: '<!-- --><script>alert(1)</script>' },
```

Everything else — the mixed-case and leading-space `javascript:`, `vbscript:`, `data:`, protocol-relative, `file://`, traversal, `<svg onload>`, `<style>`, entity storms, a bare URL in three places, an unclosed fence — is already covered there, and #168's assertions independently arrived at the same technique mine did: collect every emitted tag and check it against an allow-list rather than searching the output for scary substrings.

PR #181, opened 2026-09-10
feat(chat): an empty box asks the other model the same thing

oleksandrdubyna88:
Entry 24, and the last story of `todo/PLAN_presets_above_the_composer.md`. In the operator's words: *"maybe I don't like gemini's answer and want to switch to Fable. If I switch the model and press Enter with an empty box — take the previous context (except the last answer) and feed it to the new model."*

## What it does

**Except the last answer**, and that condition is theirs and a sound one: an answer somebody rejected, handed to the next model, is a model being asked to agree with it. The **question** is kept and re-sent verbatim — it is what they want answered again — and everything before it stays, because that is the conversation the answer was given in.

Both the answer and its question leave the transcript: the question comes back as this turn's own, and keeping the answer would show it twice in a conversation that has moved past it.

**On offer only when the model has changed** since that answer. Pressing Enter on an empty box with the same model chosen does what it has always done, which is nothing.

## The gesture is invisible, so the button says it

`send()` has always refused an empty box — which is exactly what made the gesture free to take — and a feature whose only trigger is pressing Enter on nothing is a feature nobody discovers. The Send button reads **Re-ask · &lt;model&gt;** whenever there is something to re-ask, which is both the second way in and the only way to know the first exists. Text in the box is a question and is never swallowed by it.

## It is a turn, not a second path

It goes through `oneTurn`, so the lock, the turn number a stop can name, the transcript, the model recorded on the answer and the cap on a forgetful conversation all happen because nothing about it is special.

## One of my own patches landed in the wrong place

`const box = document.getElementById('say')` appears more than once in the page script, and the push-handling block went into `fitComposer`, where `data` does not exist. Six tests said `ReferenceError: data is not defined` before it was moved — the harness that RUNS the page rather than reading it, earning its keep again.

## Tests

**1259, 1257 passing, 0 failing.**

With this the presets plan is complete: the two lists and their migration, the two rows of buttons, the tab, and the re-ask.


oleksandrdubyna88:
Superseded by #191. Everything on this branch except `src_vs_code/src/test/chatPresets.test.ts` had already landed on `main` through #179, #182, #183 and #185, and the branch had fallen far enough behind that merging it would have reverted `chatSpend.ts`, the provider split in `chatUsage.ts` and six other files — 365 deletions against 129 insertions. The seven `reaskFrom` tests SonarCloud asked for are carried across by themselves in #191, where the suite reads 1347 with none failing.

PR #187, opened 2026-09-10
fix(chat): the chat is correct on the side it actually runs on

oleksandrdubyna88:
## The symptom

Open a workspace through Remote-WSL, select text in Claude Code's panel, press `Ctrl+Alt+A`. Nothing is captured:

> copying the selection needs Windows — copy it yourself, then use "Chat with other AI" from the right-click menu

That sentence is false on that machine. The VS Code window IS a Windows Electron window; only the extension host is Linux, and `powershell.exe` is one interop hop away.

Pulling the thread found three more defects with the same shape — code judging **the machine** by `process.platform`, which describes only **the extension host**. The repository already stated that rule twice in prose (`panelProvider.ts`, `vendorTerminal.ts`) and `localEngines.ts` already obeyed it; four places did not.

| | Where | What happened |
|---|---|---|
| A | `selectionCapture.ts` | `platform !== 'win32'` refused the capture in any WSL window |
| A2 | `selectionCapture.ts` | `ran()` resolved `void`, so a helper that never started, one that hung, one that exited badly and one that ran over an empty selection all said *"select the text first"* |
| B | `chatCommand.ts` ×3, `extension.ts` | `vendors` read off the SHARED configuration. Those rows carry `executablePath`, so with *Separate settings for each side* on a WSL window launched another side's CLI — and `extension.ts` feeds `coai-mcp`, so the **gate** ran its reviewers off them too |
| C | `chatOrphans.ts` | `process.platform !== 'win32'` returned `unknown`, the one outcome that does NOT settle a row — a WSL orphan was never ended and was retried at every activation for ever |

## What was measured first

- The WSL extension host (pid 1144, `~/.vscode-server/bin/.../node`) carries `WSL_INTEROP` and **33** Windows `PATH` entries including `WindowsPowerShell/v1.0`; `execvp` from exactly that environment resolves `powershell.exe`. The whole helper runs in **1.07 s** (1080/1059/1069 ms) from `cwd=/tmp`, against **1.4–1.9 s** from `/mnt/c` and a 6 s cap — so `pressCopy`, its `cwd` and the progress label are deliberately unchanged, and the comments now say so.
- A `/proc/<pid>` ctime gave `1789026765000` for a real start of `1789026765179`, which **refutes** the reason the gate gave for reading field 22 instead. Field 22 was adopted anyway, because it is written once at fork.
- In this WSL, `claude` and `agy` are native ELF binaries while `codex` and `gemini` are `#!/bin/sh` shims **in the Windows npm directory** — so `argv[0]` is the interpreter and the recorded name sits at `argv[1]`. That is also why B is not hypothetical here.

## What changed

- **`hostSide.ts`** names what `process.platform` cannot say: `hostPlatform()` (one narrowing, replacing two identical private copies and one wrong one) and `WindowsReach` = `direct | interop | none`. Three values rather than a boolean because the person sees the difference. Nothing probes; `wslNetwork.ts` is imported from and otherwise untouched.
- **The capture** takes a reach, and `ran()` carries a `RunPhase`. Four endings, four sentences, each with what the system itself said. The clipboard is still asked FIRST: a helper that copied the passage and then exited badly has still copied the passage.
- **`sideConfig.readerFor`** is the one door to the per-side reader and derives the side from the context, so no caller passes a `Side`. The panel, the chat and the server settings file all go through it; the binding is the first statement of `activate`.
- **The orphan sweep** gained a POSIX half with both sides of the world injected — ten branches, all tested on a Windows machine with no `/proc`. One pid, not a tree, because `needsShell` gates the shim case to `win32`.

## Test plan

- `npm run typecheck` — clean.
- `npm test` — **1388 passed, 0 failed, 1 skipped.**
- Every fix landed with a test watched RED first against the unfixed code, with the real symptom in the failure: the WSL capture failed with the literal *"copying the selection needs Windows"* where the passage should have been; the settings guards failed naming three `.get('vendors')` in the chat and one in `extension.ts`; the sweep guard failed naming `process.platform !== 'win32'`.
- `node .claude/rules/shared/tools/plan-lifecycle.mjs` — clean.
- **Owed, and named as a tail:** one press in a real Remote-WSL window. No test can answer it.

## The gate

Two rounds, 15 + 30 findings. 23 taken, 20 refused with reasons recorded on the round — including a pre-flight probe (twice), a fallback for a tree walk this change does not implement, and a Blocking claim that `powershell.exe` cannot run with a Linux `cwd`, which the measurement at the top of this description had already refuted. The rounds are also where the identity rule tightened from *any `cmdline` entry* to `argv[0]`/`argv[1]` (the loose rule would kill `python job.py codex` on a recycled pid) and where a start time that could not be computed stopped reading as *not ours*.


oleksandrdubyna88:
Rebased onto `main` and taken over from the session that opened it (it had been quiet for ninety minutes, and the operator asked for this one merged and released).

**Three things happened in the rebase, and one of them was not mechanical.**

**The help conflicted in all five languages, and both halves had to survive.** This branch rewrites the first sentence of `whatCanGoWrong` — *copying the selection needs a Windows session: this machine's own, or, from a WSL window, the one that window is attached to* — while #190 rewrote that whole field and appended two paragraphs (the picture refusals, the tilde on the running total). Taking either side wholesale would have silently dropped the other. Resolved field by field: main's text everywhere, with this branch's opening spliced onto it at the stretch the two versions still share. Verified per language that both markers survive — the WSL sentence AND the paragraphs — in `en`, `ru`, `uk`, `de`, `es`.

**`todo/README.md`** — both sides added rows to *Promoted*; both kept.

**The conventions pin went stale twice while this was open.** `026b20e` landed in #189, then `895b24c` (#17, *migrate shared rules with bounded native-agent verification*) landed during the rebase. The second bump is folded in here as its own commit rather than sent separately, because a separate pin PR would put this branch behind again the moment it merged. `common/coai-review-gate.md` is byte-identical across that change, and `snippetVersion.test.ts` passes **14/14 with nothing skipped** — the mounted rule really was read and really did match.

## Verified on the rebased branch

- `npx tsc -p . --noEmit` — clean
- `rm -rf out && npm test` — **1396 tests, 1395 pass, 1 skipped, 0 failing**
- `plan-lifecycle.mjs` — only the four submodule-relative link artifacts a worktree always shows
- `pin-check.mjs` — *OK — 1 pin(s) at their remote tips*

The tail this PR named for itself stands unchanged and is still owed: **one press in a real Remote-WSL window.** No test answers it, and the operator is about to test the release.

PR #195, opened 2026-09-10
fix(claude): a session gets the instructions again, through the one door left open

oleksandrdubyna88:
## What I found

The shared layout is one source for two hosts, and it is **enforced**. `validateInstructions` in `.agents/conventions/tools/lib/rule-cli.mjs`:

- `CLAUDE.md must contain only @AGENTS.md`
- `Legacy .claude/rules is nonempty; move policy to .agents/rules before claiming one source`

I hit both by trying them, not by reading them. And those two are exactly the doors Claude Code loads project instructions through **on its own** — no imports needed, the host does it.

Measured on this repository:

| | before | after #192 |
|---|---|---|
| `CLAUDE.md` | 3 971 B | **12 B** |
| `.claude/rules/common/` loaded by the host | 6 590 B | **0** |
| a session holds, having done nothing | ≥ 10.5 KB | **561 B** |

Those 561 bytes are an instruction to go and read the rules. So the guarantee moved from the **host** to the **model's diligence** — and ENTRY.md's step 8 says it must be re-done after every compaction. The two rules that suffer most from becoming optional are this project's own: `vendor-routing`, whose violation is *invisible in the output* and has already cost three cells of a measurement, and `review-gate`, the contract for calling this product's own gate.

Content itself is fine, and I checked: local rule bodies are byte-identical (only frontmatter added), the submodule pin is unchanged, and `.agents/PROJECT.md` preserves the old `CLAUDE.md` word for word except *(Claude Code)* → *(Claude Code or Codex)*. **Nothing was lost. What was lost is the loading.**

## What this does

Leaves both enforced doors shut and uses the third — `SessionStart` — which is the only mechanism the resolver does not police. The hook runs the **same canonical resolver Codex is told to run, at the same pin**, and prints what applies to every task: `common.coding-style`, `common.knowledge-base`, `common.security`, `local.common.review-gate`, `local.common.vendor-routing` — 24 KB with BEGIN/END boundaries and hashes intact.

It does **not** replace ENTRY.md's procedure and says so in its own output: a hook cannot know at session start which files the session will touch, so `explain`/`read` for the real task is still owed. Without the submodule it reports loading **INCOMPLETE** in ENTRY.md's own words, names the two commands that fix it, and **exits 0** — a hook that fails a fresh clone is a hook somebody deletes.

## RED first, all three

- hook unwired → *no SessionStart hook runs .claude/hooks/load-instructions.mjs*
- `.claude/rules` repopulated → the door test failed **and so did the emission test**, because the resolver refuses that repository outright. That is the cleanest demonstration available that the obvious fix is not on the table.
- Green after restoring. **Suite: 1398 tests, 0 failing.** `rules check` → `resolved`.

## Next, and why PROJECT.md says so

This is the reference implementation. It is being promoted into the family reference (`dew_flow_conventions/settings/`) next, after which `.claude/settings.json` goes back to byte-identical. Until then PROJECT.md records the one addition rather than leaving its *never edit it independently* line false.

## Summary by CodeRabbit

* **New Features**
  * Claude Code now automatically loads project instructions when a session starts.
  * Instructions include universal guidance and directions for locating file-specific rules.

* **Bug Fixes**
  * Improved recovery messaging when instruction components are unavailable, including setup commands and an `INCOMPLETE` status.

* **Tests**
  * Added automated checks to verify instruction loading, configuration, rule compatibility, and graceful handling of missing components.

### DIFF
--- a/.agents/PROJECT.md
+++ b/.agents/PROJECT.md
@@ -60,5 +60,17 @@ node .agents/conventions/tools/pin-check.mjs
   entry, read once at startup via `creds config <key>`.
 - **Logging** per `.agents/conventions/common/logging-serilog.md`: coloured ANSI console (stderr in
   stdio mode) + one file per run under `logs/{yyyy-MM-dd}/`, everything UTC.
-- `.claude/settings.json` is a byte-identical copy of the family reference
-  (`.agents/conventions/settings/settings.json`) — never edit it independently.
+- `.claude/settings.json` was a byte-identical copy of the family reference
+  (`.agents/conventions/settings/settings.json`). It now carries ONE addition on top of it: a
+  `SessionStart` hook running `.claude/hooks/load-instructions.mjs`. Everything else stays
+  byte-identical, and the addition is temporary — it is being promoted into the family
+  reference, and when the pin carrying it lands here the copy goes back to being byte-identical.
+- **The Claude host adapter loads instructions through a hook, and that is not a preference.**
+  `validateInstructions` in `.agents/conventions/tools/lib/rule-cli.mjs` refuses a `CLAUDE.md`
+  that is anything but `@AGENTS.md` and refuses a non-empty `.claude/rules` — the two doors
+  Claude Code loads project instructions through on its own. Closing them is what makes one
+  source for two hosts true, and it also left a session holding 561 bytes telling it to go and
+  read the rules. The hook runs the same canonical resolver Codex is told to run, at the same
+  pin, and prints what applies to every task. Restoring instructions the obvious way instead —
+  imports in `CLAUDE.md`, or files under `.claude/rules` — makes `rules check` report
+  INCOMPLETE, which by ENTRY.md blocks edits. `claudeAdapter.test.mjs` guards both halves.
--- a/.agents/PROJECT.md
+++ b/.agents/PROJECT.md
@@ -60,11 +60,9 @@ node .agents/conventions/tools/pin-check.mjs
   entry, read once at startup via `creds config <key>`.
 - **Logging** per `.agents/conventions/common/logging-serilog.md`: coloured ANSI console (stderr in
   stdio mode) + one file per run under `logs/{yyyy-MM-dd}/`, everything UTC.
-- `.claude/settings.json` was a byte-identical copy of the family reference
-  (`.agents/conventions/settings/settings.json`). It now carries ONE addition on top of it: a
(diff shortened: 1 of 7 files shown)


oleksandrdubyna88:
fix(claude): the hook is addressed absolutely, so a session outside the root still finds it

CodeRabbit's finding, and a real defect rather than a style note. The wiring said
`node .claude/hooks/load-instructions.mjs` — a relative path, resolved against whatever
working directory the hook process happens to get. Claude Code exports CLAUDE_PROJECT_DIR
for exactly this and documents the placeholder for scripts under .claude/hooks/.

The failure mode is the bad kind: a session started anywhere but the project root dies with
"cannot find module", and the careful degradation this hook has — INCOMPLETE in the entry's
own words, the two commands that fix it, exit 0 — never runs, because node cannot load the
file that contains it. The one case the hook was written to survive is the one the relative
path breaks.

Exec form, which the same documentation recommends because it spawns the script directly
and takes shell quoting out of the picture:

  "command": "node",
  "args": ["${CLAUDE_PROJECT_DIR}/.claude/hooks/load-instructions.mjs"]

The test compares command AND args now, and asserts the absolute form specifically. RED
checked: reverted to the relative path and it fails naming it — "no SessionStart hook runs
.claude/hooks/load-instructions.mjs by an absolute path".

The same fix is in the family reference (dew_flow_conventions#19), so this stays a
byte-identical copy once that pin lands.

Suite: 1398, 0 failing.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

oleksandrdubyna88:
test(claude): assert the exec-form contract exactly, not a substring of it

CodeRabbit's second finding, and right for the same reason its first one was. Joining
command and args and looking for a substring accepts a different command that happens to
mention the path, and accepts the right one with extra arguments appended — which is the
opposite of what exec form is for. The whole point of that form is that what runs is exactly
this and nothing a shell got to reinterpret.

Asserted as a pair now: command is `node`, args are exactly
["${CLAUDE_PROJECT_DIR}/.claude/hooks/load-instructions.mjs"], compared with deepEqual so a
failure prints what was actually declared.

RED checked on the new failure mode rather than the old one: appended `--extra` to args, and
the test fails naming the contract. The previous substring version would have passed that.

Suite 1398, 0 failing. adapter-check OK.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

PR #198, opened 2026-09-11
fix(presets): Add a model wrote a row its own reader refuses, and two prompts could read as main

oleksandrdubyna88:
Two defects in the tab **Edit chat presets** opens, both reported 2026-09-11 with screenshots.

## 1. *Add a model* appeared to do nothing

It was not doing nothing. It appended `{ name: 'New model', provider: '', model: '' }` to `coai.chatModelPresets`, and `chatModelPresetsFrom` refuses a row with no provider — *"a row without a PROVIDER is not one: the provider is what answers"*. So the page re-read the list, the reader discarded the row, and the screen was unchanged. `rowsOf` reads the **raw** setting, so every press left a dead entry that no surface could display, edit or remove.

The prompt button never had the defect: its seed is two real values. The model branch seeded the one field its own reader refuses on.

**The fix is the seed, not a loosened reader.** `freshPromptRow` / `freshModelRow(taken, providerId)` live beside the readers now — the provider is a **required** argument, so a caller who forgets it does not compile. `addModel` resolves the answering row first, and with nothing that can chat it says so instead of writing something invisible. The litter is cleared when the tab opens.

## 2. Two prompts could both read as *main*

What was **saved** was right all along — ticking one unticks the others before the write. But an edit deliberately does not repaint, because a repaint under a caret is the defect the sidebar's own prompt box had, and that rule had swallowed the one edit whose entire effect is on the rows it is **not** in. A checkbox has no caret, so it is redrawn now.

## What the gate found in the fix

Plan round: 11 findings, 4 taken — chiefly that clearing the litter *"on the next write"* clears nothing for somebody who opens the tab and closes it, and nothing at all when the add refuses before writing. Code round, two passes, 23 findings, 18 taken:

- The prune **raced the first render** — awaited before the panel is created now, and its failure logged rather than left floating.
- A `main` edit **naming a row that is not there** rewrote every flag to `false`, leaving no tick while `mainPrompt` quietly fell back to the first prompt.
- The cleanup was using the **reader's acceptance rule as migration authority** — an older build opening a newer file would have deleted rows it merely does not understand. `deadModelRow` is the defect's own signature instead.
- A `main` edit is guarded by the **list** as well as the field.
- Re-ticking the prompt that is already main wrote the file back to say what it said.
- `repaintsAfter` narrowed to `editRepaints`: a whole-command policy the host consulted for edits only was a rule with a test and no effect.

## Tests

RED first on every behaviour, with the real symptom — `the row a press of Add a model writes is dropped by the reader`, and the `main` tick reported as not redrawn. Both original fixes were then proved with teeth by removing them again: the seed back to `''`, the repaint exception back out — both red, both green on restore.

`npm test` **1445 pass, 0 fail** (1427 before this branch). `typecheck`, `plan-lifecycle`, `pin-check` clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - New model presets now open with a compatible provider selected, or clearly indicate when none is available.
  - Unusable model rows left by earlier behavior are automatically cleaned up.
  - Only one prompt can be marked as the main prompt at a time.
  - Main-prompt changes update immediately, while unrelated edits avoid unnecessary refreshing.

- **Documentation**
  - Added documentation describing the preset fixes and resulting behavior.

### DIFF
--- a/research/module_extension.md
+++ b/research/module_extension.md
@@ -860,6 +860,23 @@ marked — refusing it would mean the silent substitution this whole change remo
 cannot be written says so once a session rather than once a render, because the consequence is one a
 person can act on: the section goes on offering models the command will not find.
 
+**Two defects in the CRUD tab, reported 2026-09-11 with screenshots.** *Add a model* appended
+`{ name: ‘New model’, provider: ‘’ }` and `chatModelPresetsFrom` refuses a row with no provider — so
+the page re-read the list, the reader dropped the row, and the screen was unchanged while
+`settings.json` grew a dead entry per press. The seed moved next to its reader as `freshPreset`,
+which is the whole point of where it lives now: a shape its own reader refuses cannot be written
+there without a test going red. `addModel` resolves the answering row FIRST — the first that can
+chat, the same default the chat itself takes — and refuses by name when none can, because a button
+that silently writes litter is worse than one that says why it cannot. `readableModelRow` clears the
+rows the old behaviour left, which are not drafts: no surface has ever been able to show one.
+
+And two prompts could both read as **main**. What was SAVED was right — `edited` unticks the others
+before the write — but an edit deliberately does not repaint, because a repaint under a caret is the
+defect the sidebar’s own prompt box had, and that rule had swallowed the one edit whose entire effect
+is on the rows it is NOT in. `repaintsAfter` carves the exception, beside `presetEdit` and tested with
+it: a checkbox has no caret. Unticking the siblings in the page’s own script was the alternative and
+was rejected — the same rule written twice, and the second copy is the one that drifts.
+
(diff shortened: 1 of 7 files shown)


oleksandrdubyna88:
All three taken.

**The stale id** — `edited` mapped over the rows and produced a new array with identical content, so the host's reference check read it as a change and wrote the file back. Fixed, and it moved while it was being fixed: `editedRows` now sits in `chatPresetsPage.ts` beside `presetEdit` and `editRepaints`, which is where the command-shaped decisions live. The list-shaped rules stay in `chatPresets.ts`.

**The head-only assertion** — you are right that `dead-2` could survive the write and the test would still pass, because the reader filters it out again on the way back and hides exactly what the test is for. It asserts the whole written list now.

**The changelog** — it described the first draft, which cleared the rows on the next write. It says "when the tab opens" now, which is what the code does.

`npm test` 1446 pass, 0 fail.

PR #207, opened 2026-09-11
feat(chat): two ways to ask, a tab that knows its session, and a button that remembers what you asked

oleksandrdubyna88:
Three things the operator asked for, in one branch.

## 1. Two menu items that say what they do

The right-click menu offered one **Chat with other AI**, and what it did depended on
`coai.chatAutoSend` — a setting in another window. An item whose behaviour you cannot predict from
its own label is one you have to remember a setting to use.

- **Chat with other AI: default** — the model ticked as main, asked straight away.
- **Chat with other AI: choose** — the turn goes into the composer; change the model or the prompt
  and press Enter yourself.

Neither reads the setting. The chord and the palette keep the setting-dependent command, where the
name promises nothing about what happens next. Both items appear in both menus — the Claude Code
panel and an ordinary editor.

## 2. A tab knows its own session

Asked as *"по айди окна разве не можем определить однозначно?"* — by window id, no: VS Code exposes
none. But Claude Code writes `{"type":"ai-title","aiTitle":"…"}` into its session file as it names
the conversation, and that string is exactly what VS Code puts on the tab. Measured against a live
session before this was built.

So two sessions waiting in one folder are no longer a flat refusal. Everything genuinely ambiguous
still refuses, by name: a title naming neither, a title naming both, no tab at all — and, added by
the gate, two workspace **roots** each holding a session by that name.

## 3. Asked — the button that remembers what you asked

> *"когда окно долго работает (4 ч и более) начальный вопрос исчезает, и мне приходится спрашивать
> клод над чем ты работаешь"*

A chat tab open for hours has lost the thing it is about. The words never went anywhere — Claude Code
appends every one of them to its own session file.

**Asked** sits in the header beside the ± steppers, where it was asked for. Pressed, it opens a
region **above** the conversation and **outside** the scrolling one, so scrolling cannot move it
(*"неважно где я, в начале или в конце"*). `‹` and `›` step through every turn the person typed,
oldest first — their answered choice when asked whether one was enough. Pressed again it folds away
over half a second, not at once. A tab opened from a **file** has no session behind it, so it has no
button at all.

**What counts as the person speaking**: `origin.kind === 'human'` on a `user` row, minus a
`tool_result`, a sidechain, a meta row and a local command's own stdout. A turn an extension
*prefilled* is still theirs — CredsForDevs writes its preamble into the composer and the person types
at the end of it, so the two arrive as one message, confirmed with the operator rather than guessed.
Every `text` block of a turn is joined. A `<command-name>` envelope is unwrapped only as a contiguous
block at the start, and it already carries its own slash.

## What the gate changed

Two rounds, fifteen vendors' reviewers. The plan round passed `good_enough`, the code round `revise`
then `proceed`. Twenty-one findings taken.

- **The read is streamed.** `promptsInSession` read every candidate session file into a string and
  split it into an array of lines — to compare a title. Five reviewers across two vendors measured
  the same shape at 10× as seconds of a blocked extension host. Two passes with `node:readline` now:
  titles over every candidate, prompts over the one that matched, stopping at 200 turns, each cut at
  8 000 characters saying where it was cut.
- **A tab pins its FILE, not the name that found it.** A title moves — Claude Code refines it and the
  tab follows — so a name captured at open stops matching by the afternoon, which is exactly the
  window this feature is for. Resolved in the background as the tab opens and kept; only ever when
  exactly one session across every root matches.
- **The seam counts presses.** Open, fold, open again starts a second read while the first is
  running, and the slower one landing last would put back what the person had moved on from.
- **Every outcome is named.** No session file, no folder open, a namesake, a conversation nobody has
  spoken in yet, and a pinned file that has gone all look identical from an empty region.
- **A stored tab that does not say which door it came through is not a session.** The plan argued the
  opposite and was wrong: a file chat called `README.md` restored beside a session Claude named
  `README.md` would have shown that session's words inside the file's tab. The reversal is recorded.

## What CodeRabbit found, after the gate

Four more, two of them real defects nobody else had seen.

- A **restored tab resolved its session and threw the answer away**, so every later press searched by
  a name exactly as stale as the first — a conversation Claude renamed after a reload could never be
  found again.
- The **folded region was folded only to the eye**: `max-height` and `opacity` leave the arrows in
  the keyboard tab order and the text on a screen reader. It carries `visibility` now, which animates
  discretely and is therefore exactly right — visible the instant it opens, hidden once the fold
  ends — plus `aria-expanded`, `aria-controls` and `aria-hidden`.
- `[^<]*` **refused a person's own angle brackets** in command arguments, so a prompt asking about
  `Array<T>` came back wearing its tags.
- The **help promised the keybinding always asks at once**, while `triggerPlan` consults **When to
  send** whenever no menu item answered for it. Corrected in all five languages.

Its finding about pinning while another workspace root is ambiguous was already fixed, found by
re-reading the diff against the repo's own rules before the review landed. Three findings that the
dates are in the future were declined: this machine reads 2026-09-12 on both clocks.

## What the operator's own check found

Six defects, after two gate rounds, fifteen reviewers and CodeRabbit. Every one needed a real
machine, and the deciding one had been quoted back in the refusal on the very first screenshot.

1. **The tab wears a SHORTENED name.** Claude Code truncates the title on its own panel and writes
   the whole of it to the session file, so an exact comparison matched short conversations and never
   long ones. A name ending in an ellipsis is a prefix now; two sessions fitting one prefix are the
   same refusal two sharing a whole name always were.
2. **A window with no folder open** was told there was nowhere to look. Claude Code runs without
   one, in the home directory. (Reading every project instead was tried first and was worse than the
   bug: 77 directories, 1.0 GB, still reading when the screenshot was taken.)
3. **`hidden` did nothing** on the row of arrows — `.askingHead { display: flex }` outranks the
   browser's own `[hidden] { display: none }`. The test asserted the property on a stub and never the
   stylesheet, so it was green while two dead boxes sat on screen. It reads the CSS now.
4. **Take the question had the same no-folder bug**, and said so out loud: *"Open a folder first — a
   Claude Code session belongs to one."* It does not. That command had never once worked for an
   operator who works without a folder. `foldersToSearch` is one decision now, in the pure module,
   because it was wrong in two readers at once precisely by being written twice.
5. **One waiting session went to whichever tab asked.** The title was consulted only to break a tie
   between two — backwards, since a tie is where a wrong pick is suspected and a single candidate is
   where nobody would ever know.
6. **A question you INTERRUPTED counted as answered.** Claude Code writes a rejection as a
   `tool_result` exactly as it writes an answer; the difference is `is_error: true`. Both shapes were
   read off the operator's own file. So a question sitting on screen with its Submit button was
   reported as settled, and another conversation's question offered instead.

Also from using it: the three menu items are named **CoAI: default** / **CoAI: choose** / **CoAI:
take the question**, and the line under the pinned region is orange, because in the panel border's
own colour it read as one more division among many.

## Release

**Extension 0.36.0**, in this PR as 0.35.0 was in its own. The changelog entry is written for
somebody deciding whether to update; `npm run package` produces the VSIX at 0.36.0.

## Test plan

- [x] `npm run typecheck` clean.
- [x] 1618 tests, 1617 pass, one skipped; the VSIX packages at 0.36.5; `plan-lifecycle` clean.
- [x] The button is pressed in the **shipped, minified bundle** — the region, the arrows and the
      answering message are four more names a minifier gets to rewrite, and the rounds log has been
      broken exactly that way twice. Watched red with the class renamed: *"the shipped page did not
      open the region"*.
- [x] Help updated in English, Russian, Ukrainian, German and Spanish in the same commits.
- [x] `research/module_extension.md` updated; the plan promoted to `research/` with its deviations
      and its open tail; `plan-lifecycle` clean.

## Summary by CodeRabbit

- **New Features**
  - Added separate **CoAI: default** and **CoAI: choose** actions for sending questions to another AI.
  - Added **CoAI: add the question** to append captured questions to existing composer text.
  - Added an **Asked** panel for viewing and navigating questions from the originating Claude Code session.
  - Chat tabs now match and retain their associated Claude Code session, including sessions in folderless windows.

- **Documentation**
  - Updated help content and changelog entries across supported languages.


oleksandrdubyna88:
Taken four of these, and thank you — two of them were real defects nobody else had seen.

**Fixed in 5a995df:**
- the restored tab that resolved its session and never kept it, so every later press searched by an equally stale name;
- the folded region that was folded only to the eye — it carries `visibility` now, plus `aria-expanded`, `aria-controls` and `aria-hidden`, all kept in step by `showAsked`;
- `[^<]*` refusing a person's own angle brackets in command arguments;
- the help in all five languages, which promised the keybinding always asks at once while `triggerPlan` consults **When to send** whenever no menu item answered for it.

**Already fixed, in 701cb18**, pushed before this review landed: pinning a file while another workspace root reported namesakes. Found by re-reading the diff against the repo's own rules — two roads to the same defect, which is a good sign for both.

**Declined — the three date findings.** They are not in the future: this machine reads `2026-09-12` on both the UTC clock and the local one, and the commits are timestamped `+0200`, so a reader in UTC sees the first of them at `2026-09-11T22:01Z` and infers a day that has already passed here. The Definition of Done you flagged beside them was a fair catch and is ticked in the same commit.

oleksandrdubyna88:
Second pass — all three taken, in `ac107c8`.

**The stale clause was the best catch of the two rounds.** English and Russian had that paragraph rewritten whole; Ukrainian, German and Spanish had only their *first* sentence replaced, so each went on saying the menu item "puts it in the box and YOU send it" — true of `CoAI: choose`, false of `CoAI: default`. The sentence I replaced was fine; the one after it was the one that had gone stale. A good argument for rewriting a paragraph rather than patching its opening.

**The raw reason** is gone from the toast, in *both* registrations — the pre-existing one had the same shape and the path rule applies to it equally. The detail goes to the console, because a command that fails and explains nothing is the defect the `catch` was added for in the first place.

**The orange rule** now falls back through `--vscode-editorWarning-foreground` before reaching the hex.

The three date findings stay declined: this machine reads `2026-09-12` on both the UTC and the local clock, and the commits are stamped `+0200`, so a reader in UTC sees them a day earlier and infers a date that has already passed here. The Definition of Done flagged beside them was a fair catch and was ticked earlier in the branch.

Resolving these threads now — every one has been acted on or answered above.

PR #216, opened 2026-09-12
feat(roles): a page where a person writes their own review role

oleksandrdubyna88:
Plan 2 of 5 on making review ROLES user-definable. `coai-mcp 0.19.0` reads `COAI_ROLES` and **nothing wrote it** — somebody who wanted the gate to check a CV, a product description or a set of requirements had to hand-write JSON into `settings.json` and hand-place a markdown file beside it.

**Edit roles…** in *Prompts per round* now opens a page. Add a role, name it in your own language, write the question it asks, add more prompts for narrower questions. The five shipped roles are on the same page with their prompt text editable. A prompt's body is a FILE, at `<dataDir>/prompts/<id>.md`, which is where `RolePrompts` has read overrides since before roles were data.

Plan: [todo/PLAN_review_roles_crud_tab.md](todo/PLAN_review_roles_crud_tab.md). Follows [PLAN_review_roles_become_data.md](research/PLAN_review_roles_become_data.md), shipped as `mcp-v0.19.0`.

## What the gate found

Two code rounds: **34 gating findings, then 9.** The two worth the first round:

- **`composed()` disagreed with the server about what a row may say.** It applied a row's `name`, `stage` and `programmingTask` to a role this product ships; `RoleComposition.Overridden` takes only `Active` and the extra prompts — *"Id, name, stage and kind are NEVER taken from the row."* Nothing failed, which is why it survived a full suite: the page drew the person's word for the role and every round used the shipped one, and the only place the two met was a log nobody reads beside a panel nobody doubted.
- **"Remove this role" did nothing**, for the whole plan. The host's mutation returned `RoleRow | undefined`, where `undefined` meant BOTH "refused" and "nothing changed" — so removal, which produces no row by definition, took the refusal's behaviour and wrote nothing.

**`rolesEdit.ts` is new, and is the structural answer to how that survived.** The rule lived inside a webview host, which this extension deliberately does not unit-test, so no test could have caught it and four reviewers found it by reading. Every row decision moved into a pure module returning a three-way union — `rows` (carrying the prompt-override files that no row points at any more), `refused` (carrying the sentence to show), `unchanged`.

And the best of the second round: **a stage's rules were enforced on one of the three doors into it.** How many roles are active in a stage is changed by the switch, by removal, and by a stage MOVE — only the switch was checked, so the page would save an empty stage or a sixth active role by either of the other two routes.

Also fixed: `enabledCodeRoles` is the ONE count of which code roles will run (there were three, and the fan-out sentence walked a fourth list); `coai.roles` is per-side and was being written globally while the page said "saved for this side of the machine", so `sideConfig.saveSetting` is now the single write path shared with the panel; `composed()` returned two shapes and callers were compensating with `?? true`; the sidebar's tick is inert for a role the catalog has switched off, because it wrote the other switch and sprang back silently; commands are serialized and typed fields settle rather than writing per keystroke, and a typed field no longer repaints under the caret; prompt bodies are written beside their destination and renamed over it; removing a role asks first, then deletes its prompt files; a prompt id from the webview must belong to the role that claims it; `rolesFrom` carries fields it does not know rather than deleting them on the next keystroke — which is how `remoteVendor` was lost by three releases; generated ids are length-capped because they become environment variables; and the nonce comes from `crypto`.

26 findings were rejected with reasons, the largest group being four claims of script injection through `JSON.stringify` inside a `<script>` — there is none in the file, and no page state reaches the script block at all.

## Test plan

- [x] Extension suite: **1832 passed / 1 skipped / 0 failed** (`out/` removed first)
- [x] `CoaiMcp.Tests`: **1298 passed / 1 skipped / 0 failed**
- [x] `CoaiServer.Tests`: **245 passed / 0 failed**
- [x] `plan-lifecycle` clean, `adapter-check` clean
- [x] Every defect watched RED first, with the failure naming the real symptom — the built-in rename returned `'Mine'` where `'Plan review'` was expected, removal returned `{"kind":"unchanged"}`, the unknown field came back `undefined`, the generated id was 480 characters
- [x] `composed()`, `promptBelongsTo`, `lastStanding` and `materialised` each broken again after the fix and caught by name
- [ ] **Not verified end to end against an installed `coai-mcp`** — a prompt body written by this page and read by a real round. That is the one Definition-of-Done line this PR cannot tick from a test.

## Notes

- `pin-check` reports the conventions pin stale. **This branch's pin is byte-identical to `origin/main`'s** — another session's conventions commit landed after main's last bump, so it is not this PR's change to make.
- Two things were declined as decisions rather than defects and are recorded in the plan's open tail: prompt bodies are global while role rows are per-side, so two sides creating a role of the same name share one body file (changing that changes what `RolePrompts` reads — a wire change, and an operator's call); and `stage !== PLAN_STAGE` is still what "a code role" means, which plan 4 has to replace when a third stage exists.


oleksandrdubyna88:
@coderabbitai review

PR #223, opened 2026-09-13
feat(chat): conversations live on disk, not in the editor's state

oleksandrdubyna88:
Epic A of `todo/PLAN_go_to_conversation.md`: conversations move out of VS Code's `workspaceState` and into a file store, so a person can reopen one two days later.

**Nothing in this branch is visible in a running window.** It is the machinery under the floor. The commands, the picker and the new-chat button are epics B, C and D, on a later branch.

## Why the memento had to go

Chat transcripts lived in one JSON value under one key, rewritten whole on every change, cut at seven days and twenty records. The feature needs ninety days of them, and the operator named the ceiling: a memento is written to the editor's SQLite state, it lags as it grows, and it drops or truncates silently when it overflows. Ten tabs a day is past the twenty-record cap in two days.

## What the four stories built

| Story | What it is |
|---|---|
| **A1** | The record and its validation, and one shared atomic write — write beside, rename over, with a bounded retry for the transient Windows rename codes. |
| **A2** | The file protocol. One conversation is a transcript and a metadata file; the listing never opens a transcript. A save is a compare-and-swap on a revision, claimed by an exclusive create — one lock per conversation — with a fenced release and a staleness break. |
| **A3** | The dual write: every conversation written to the store beside the memento, nothing reading it. A refusal forks under a new id rather than overwriting anybody. |
| **A4** | The cut-over. The store becomes the source of truth and the memento is emptied — but only after every record is proved present, and only once. |

## What the gate found

Five plan rounds and six code rounds. Three defects would have destroyed people's conversations silently, and each was found by two vendors independently:

- **The compare-and-swap was a check-then-act.** Reading a revision and then renaming conditions nothing on what was read, so two windows both holding revision five both wrote six and the second won silently.
- **The lock's fence was a read-then-delete.** A stalled writer could remove the lock of whoever had replaced it, letting a third writer in.
- **The migration treated "the store has this id" as proof the store's copy was newer.** It is not: the dual write is best-effort, so the memento could hold eight turns where the store held five. Emptying the key would have deleted three turns permanently, with every test green. The migration compares transcripts now.

The gate was also wrong often enough to be worth checking. One round claimed every quarantine write fails for a missing directory, which the atomic write creates and an existing test already disproved; several findings argued themselves round to "this is correct" before reaching a conclusion. Every rejection carries its reason in the commit that answers it.

## Test plan

- `npm test` in `src_vs_code`: **2019 tests, 2018 passing, 1 skipped, 0 failing** on the rebased branch. The suite was 1832 before this work; the skip is pre-existing.
- Every structural guard had its teeth proved by breaking the wiring and watching it go red with the real symptom, then restoring it — the migration wait, the seal order, the store re-check, the dispose arm, the retirement gate and the old prune among them.
- The migration is tested against real temporary directories: a memento newer than the store, an interrupted run, an unavailable store, a quarantined record, a concurrent write before the clear, and a folderless window.

## Not in this branch, deliberately

The sweep and the in-memory index (B1), the picker (B2–B4), the durable source that lets a tab find its own conversation (C1), and the reset (D1–D2). Every record carries no source yet, which matches no tab — the honest answer until C1.

**One gap is recorded rather than hidden:** the memento's retention went with it and the store's own sweep is B1, so between the two stories the conversations directory grows with nothing retiring it. The plan's growth table says so, and says B1 follows within one working session or the sweep moves into A4.

No user-visible change, so no version bump and no changelog entry; the plan puts those at B3 and D2.

## Summary by CodeRabbit

- **New Features**
  - Conversations are now stored durably on disk, allowing them to persist across reloads and extension-host restarts.
  - Existing conversations are migrated automatically, with conflicting copies preserved as separate conversations.
  - Concurrent changes are handled safely, with clear notices when a conversation continues elsewhere.
  - Chat tabs can restore with retryable guidance when storage is temporarily unavailable.

- **Documentation**
  - Added architecture and module documentation covering conversation persistence, migration, synchronization, and recovery behavior.


oleksandrdubyna88:
Fixed in 24db2dd, and thank you — nine of the ten were true, and one of them led somewhere better than the comment itself.

The lock clock was the best catch. The migration threaded one instant through every read and save, and the store writes that into the lock note and ages rival locks against it — so on a slow disk a lock the migration is actively holding looks stale to another window. The fixture then demonstrated the same defect a second way: stamped ahead of the real clock, a lock taken a moment ago already read as stale.

The message validator turned out to have a second cause behind it. That rule existed twice — once for the record on disk, once for the record in the memento — which is why only one copy had been tightened. The memento's copy still accepted the malformed message, so the migration carried it to disk and the store then reported it unreadable: a conversation that should have been quarantined, where a person can go and get it, became one that reads as broken. Both now import one module.

Also taken: the disposed-panel guard, the note left standing over a state that should replace it, both exhaustive switches with a never arm, the test pinning a refusal that reports no transcript at baseline zero, and all four documentation points — the status line, the architecture sentence that overstated forking, the older module section that still called the memento authoritative, and the index that called the store one file when it is two.

One correction for the record: the ReadOutcome switch lives in chatRestore.ts rather than chatStoreWrite.ts, so the never arm went there.

PR #224, opened 2026-09-13
feat(roles): a Team server runs the roles a person wrote

oleksandrdubyna88:
Plan 3 of 5 on making review ROLES user-definable. Plans 1 and 2 made a role data and gave a person a page to write one — and **a Team server still ran only the five it was compiled with**, so a team that shares one was exactly the team that could not share a review role somebody had written.

Its operator sets `Coai:ExtraRoles` on the server — or `Coai:AllowAnyRole` to accept any role at all — and the server runs them beside the five it ships with. **A server nobody reconfigures behaves exactly as it does today.**

Plan: [todo/PLAN_team_server_accepts_custom_roles.md](todo/PLAN_team_server_accepts_custom_roles.md). Follows [PLAN_review_roles_become_data.md](research/PLAN_review_roles_become_data.md) and [PLAN_review_roles_crud_tab.md](research/PLAN_review_roles_crud_tab.md), both shipped.

## Four stories, each gated on its own diff

| | |
|---|---|
| **1** | `AcceptedRoles` — the one place that decides which roles this server runs. Building it IS the boot guard: a configured id that could never run stops the server, naming it. |
| **2** | Both role gates and the request ingress read it. A structural test fails if a third list appears. |
| **3** | `/api/catalog` carries `roles` and `allowAnyRole`, populated from that same instance. |
| **4** | Both clients ask instead of guessing, and the panel says it **before** a round rather than in its result. |

## What the gate found

Five rounds, **86 gating findings**, 48 accepted. The ones worth reading:

- **The plan itself was wrong about `AllowAnyRole`.** The first draft documented a usage-ledger split as an acceptable price and put the question to the operator. Three reviewers refused that independently and one named the answer I had dismissed — **case-fold the id**. `Requirements` and `requirements` become one row on every server with nobody configuring anything.
- **A transient failure disabled custom roles for ever.** The warm-up skipped any server whose state was no longer "not asked", so one 502 during a restart recorded *unreachable* permanently — and an operator adding to `Coai:ExtraRoles` was never heard again. Four findings from two reviewers were this.
- **The two C# copies of the role-id grammar disagreed**, for the length of one commit: in .NET, `$` matches at the end of a string **or immediately before a trailing newline**, so `^[A-Za-z][A-Za-z0-9_]*$` accepts `"Requirements\n"`. My own RED test caught it on the server; a reviewer caught that the client had not been fixed with it.
- **A chat carrying an unknown role got contradictory advice** one request apart: first "'Invented' is not a review role. Accepted: …", then — after the client picked one of those names — "a chat carries no review role". The kind is judged before the role now.
- **The two clients read a malformed catalog differently.** `{"roles":[null]}` made coai-mcp exclude every shipped role while the extension read it as "said nothing useful". One server, two clients, two different rounds.
- **Cancellation was swallowed into an unreachable verdict**, so a round whose clock ran out poisoned the cache and carried on past its own cancellation point.

## The decisions worth arguing with

- **Two wire fields, not one with a sentinel.** `["*"]` was considered and refused: `roles.includes(role)` is the obvious client code and answers `false` for every real role, so a server's whole configuration would be inverted by the one line anybody would write.
- **Absent means the five this product ships** — never "none", never "any". An empty list means the same, because a server that *has* the field always accepts at least five. This is the rule `remoteVendor` was lost twice for want of.
- **Three answers, not two.** A server that answered without the field and one that could not be reached both leave a client without a list. The fallback is identical; only one of them can honestly say why a role was left out.
- **The rule is written twice, once per client, deliberately** — two binaries that ship separately, which is the argument `RemoteCatalog` already carries.

## Test plan

- [x] `CoaiServer.Tests` **306 passed** (from 245)
- [x] `CoaiMcp.Tests` **1337 passed / 1 skipped**
- [x] Extension **2051 passed / 1 skipped**
- [x] `plan-lifecycle`, `pin-check`, `adapter-check` clean
- [x] Every defect watched RED first; the case-fold, the boot guard, the stage invariants, `composed()` and the catalog population were each broken again afterwards and caught by name
- [ ] **Not verified against the deployed server.** That deploy is manual (`workflow_dispatch`), so it is the plan's last Definition-of-Done line and a step somebody performs.

## Notes

- `Coai:ExtraRoles` and `Coai:AllowAnyRole` are documented in [research/module_team_server.md](research/module_team_server.md); [research/architecture.md](research/architecture.md) gains the discovery flow, because this changed how two clients and the server talk.
- The help article is updated in English **and all four translations in the same commit** — a translation that is behind is invisible, since `bodyFor` marks a missing one and never a stale one.

## Summary by CodeRabbit

- **New Features**
  - Team servers can run operator-configured extra review roles or accept any valid role identifier.
  - Catalogs now report the roles each server supports.
  - The panel checks configured servers before each round and identifies roles a server cannot run.

- **Bug Fixes**
  - Older or unreachable servers now fall back to the five built-in roles with a clear status message.
  - Role names are consistently validated, normalized, and matched without regard to case.

- **Documentation**
  - Updated help content and release documentation across supported languages.


oleksandrdubyna88:
All nine addressed in ebf31e6b — eight fixed, one declined with a reason.

**Fixed**

- **Role names from a catalog are held to the ID rule, in both halves.** You were right that `[' ']` made a catalog look *answered*, and an answered catalog is the whole truth about its server — so one space would have reported every shipped role as unsupported. `isRoleId` is exported from `roles.ts` now so the extension applies the same shape and the same 48 as everything else, and `RemoteRoles.From` applies it on the C# side. Both trim first, because the server trims what it is sent.
- **A stale catalog is no longer an answer about roles.** This was the best of the nine: `refreshCatalog` keeps the previous catalog and marks it `stale` so the Team servers section can say what it last knew, and "what it last knew" is a different claim from "what this server runs". It reads as unreachable now, with a regression test.
- **The missing-role refusal now knows about `AllowAny`.** Listing `Names` on a server that accepts any well-formed id presented five configured names as the accepted set — the same misdirection the *shape* refusal was split out to avoid, one branch up. I had fixed it there and missed it here.
- **Complexity.** `WhyNot` was 9 against the ceiling of 4 and `AskAsync` was 5. Split: the answered case against the three not-answered ones, and the response handling out of the try/catch.
- **Two places where my own prose contradicted my own three-state distinction.** An old server is not "too old to be asked" — it answers, without the role fields; only an unreachable one could not be asked. And `Coai:AllowAnyRole` does not "accept any role at all": it still applies the shape rule, and this plan's own server refuses `My-Role` under exactly that flag. Corrected in the CHANGELOG and in all five languages.
- **The plan document.** The status line did contradict itself; decision 3 still quoted the `^…$` form that story 1 replaced after finding that .NET's `$` matches before a trailing newline; and the canonicalisation section claimed an order the code does not have — the gates read what the client sent so a refusal can quote it back, and the canonical value is built after them for the record and the fingerprint.

**Declined**

- **Promoting the plan in this PR.** It is promoted on merge, in its own PR, as plan 2 was — `plan-lifecycle` is clean with it in `todo/` until then, and the promotion needs the merge commit to reference. The self-contradiction you found in its status line is fixed either way.

Tests: CoaiMcp 1337 passed / 1 skipped, CoaiServer 306 passed, extension 2055 passed / 1 skipped.

PR #272, opened 2026-09-15
feat(store): the accepted findings read back as a corpus

oleksandrdubyna88:
## What

`Store/BugsQuery` reads the accepted findings back as **material** rather than as a log, through
`coai-mcp --bugs-json [--limit 200] [--all]`. Story 1 of `todo/PLAN_a_corpus_of_real_defects.md`
(#269).

An accepted finding is by definition a defect somebody confirmed — the blind-spot corpus this
database was built for on 2026-09-05. Nothing has ever read it back that way: `--log` answers *what
happened in this round*, and *which defects are worth keeping* filters on things it has no opinion
about.

## ⚠️ Merge order with #268

Both this and #268 append a step to `Schema.Steps`, so both are "step 5". **Whichever lands second
must be rebased** so the two steps are ordered, and neither branch's binary should be run against the
real data directory before then — a database migrated to `user_version = 5` by one of them would skip
the other's step for ever. That is why the verification below ran against a copy.

## The filter, and where the numbers come from

Six conditions, measured over a live database *before* any of this was written — 8 687 findings
narrowing to **462** usable candidates across 115 stories and 15 repositories:

| step | left |
|---|---:|
| all findings | 8 687 |
| a **code** round (a plan round's remarks are prose about a document) | 5 705 |
| resolution **accept** (a rejection is a disagreement; undecided is neither yet) | 2 622 |
| **gating** severity (a Nit accepted to be agreeable teaches nothing) | 1 917 |
| a **runtime** category (Architecture/Ux/Convention judge shape, not failure) | 771 |
| a **file and line** (or it cannot be read back out of git) | 771 |

That last step costs nothing and is kept as a guard rather than dropped.

## Decisions

- **The funnel is the shape, not a total.** `BugFunnel` carries the count after each step *and* every
  step before it. A flat "30 % were not collectable" lets the step that drops an unparsed language
  mask the step that drops the fixes nobody can find — and the second is the one with a decision
  attached to it.
- **`collect_state` is TEXT, not a flag.** A boolean says only "looked at", so the first run consumes
  every finding it saw and a second run — better prompt, wider language set, repaired walk — can
  never reach them. With `collect_reason` beside it, a skip rate becomes a measurement.
- **The two SQL texts are written out, not composed.** The first version joined an array of
  predicates so the funnel and the candidate `WHERE` could not drift. It was safe — every element a
  literal in the file — and read to any scanner as SQL assembled at runtime. This repository already
  had that argument on `RoundsQuery`'s paired statements and settled it the other way; the drift is
  held by a test instead.
- **It is the one read-only mode that migrates first.** `BugsQuery` never writes, but it is the first
  caller to want the collector's columns, and a person may run `--bugs-json` before the server has
  opened the file with a build that has them. Without that, the honest answer to an old schema is
  nought candidates — which reads as *no material* rather than *wrong schema*. A database that will
  not open at all is reported with **74**, not answered with an empty corpus.

## Test plan

- **Eight tests, red first**, against a stub returning an empty corpus — behavioural failures
  (`Expected … to contain a single item, but the collection is empty`), not compile errors.
  - the happy path, field by field;
  - `EverythingTheCorpusCannotUse_IsLeftOut` — one round carrying every exclusion at once, because
    the risk is not that each clause works but that the six compose;
  - `TheFunnelCountsEachStep_AndItsLastStepIsTheListItself` — the anti-drift guard for the two texts;
  - already-handled findings hidden by default and reachable with `--all`;
  - no database is an empty corpus, not a failure; the list is capped and the count is not;
  - the mode has its own argument, and a near miss is refused rather than guessed;
  - `ADatabaseFromBeforeTheCollector_GainsItsColumns_AndItsFindingsAreMaterial`.
- **The migration test was proven to have teeth** by removing `TheCollectorsState` from
  `Schema.Steps` and watching it report an empty collection.
- **Full suite: 1 901, 0 failed**, 2 skipped (both need symlink elevation on this machine).
- **End to end against a copy of a live 25 MB database**, never the original. The binary's funnel
  reproduces the measurement's shape on data grown by forty minutes of real use —
  9 078 / 5 980 / 2 727 / 1 983 / 791 / 791 — including `located == runtime` exactly. The copy
  migrated to `user_version` 5; the original stayed at 4.
- AOT publish: no IL2xxx/IL3xxx warnings. The native link step could not run locally (`vswhere.exe`
  is not on PATH in this shell), so the runner does that half.

## Not in this PR

Nothing writes `collect_state`, `collect_run_id`, `collect_reason` or `fix_sha` yet — the collector
does, in story 3. All four are read back on every candidate, so none of them is a column nobody
looks at.


oleksandrdubyna88:
Rebased onto main now that #268 has landed: `TheCollectorsState` is migration **step 6**, after `WhatItWasAgainst`. Inserting it ahead would leave every database already migrated by main without the collector's columns while believing it had run every step.

SonarCloud was right — 71.4 % on new code, because `BugsJson` had no test at all. Three tests now drive it end to end (the JSON, a machine that has reviewed nothing, `--all`).

That immediately surfaced a real defect the targeted runs could not see: this file and `ABatchFindingsReadTests` both replace the process-global `Console.Out` while xUnit runs their classes in parallel. One full run failed here with `'{' is invalid after a single JSON value`, the next failed *there* with `the input does not contain any JSON tokens` — the same defect from each end. Both are now in one `console-out` collection with `DisableParallelization`, the convention `fakecli-env` already set. Two consecutive full runs clean: 1 906 tests, 0 failed.

PR #280, opened 2026-09-15
chore(release): extension 0.46.0

oleksandrdubyna88:
Version bump and changelog for the extension release. **No source changes** — the gate does not apply to this diff, and the work it releases went through it on its own branches.

## What 0.46.0 carries

**The Consultant section picks a vendor the way *Add a reviewer* does** (stories C5 and C6, plus four gate rounds of fixes). The picker listed reviewer ROWS in internal ids — `codex · gpt-5.6-luna` — two sections below a picker offering `Codex (OpenAI)` and `DeepSeek`. It now reads the catalogue, and the blank preset that could not be offered before is carried as a request through the flow that adds a reviewer, rather than as a setting write.

**A copy that did not land can no longer put stale text back over what you copied next** (#277). 0.45.1 shipped the per-block copy controls; this fixes a defect in the machinery underneath them. A clipboard write cannot be cancelled, so a stalled one releases the queue and the newer text is put back when it finally lands — but that corrective write was itself started and forgotten, so a press landing while a correction was in flight let the correction settle last.

Behind it: every write and every correction take one path; the newest **press** wins rather than the newest write to *start*; and a correction goes out on the ports of whoever wanted the text. The guarantee is bounded to one extension host and says so — two windows are two hosts.

## Tests

`npm --prefix src_vs_code test` — **2917 tests, 2916 passed, 1 skipped, 0 failed.**

## After the merge

The tag `extension-v0.46.0` is pushed **alone** on `main`, and the run it starts is verified against the run ids recorded before the push — a push of several tags at once creates no workflow events at all, silently, and a green `git push` is not evidence a release began. Then `POST_DEPLOY.md` against what the Marketplace actually serves.


oleksandrdubyna88:
Closing: superseded rather than rejected. A parallel session cut the same version while this was in checks — `extension-v0.46.0` is tagged and its release published at 13:07:47Z, carrying the same commits this bump was for (#277's corrective-write fixes and the consultant work). Bumping again would put two builds behind one version string, which is the one thing a version must never do.

Nothing here is lost: the changelog entry this PR carried describes work that shipped, and 0.46.0's own entry covers it. Verified before closing that the published tag contains all three commits.

PR #304, opened 2026-09-15
fix(panel): Add a reviewer offers the whole catalogue, and names Claude Code

### LINKED ISSUE #294: add reviewer нет клод кода
Image
нужно добавить клод код что б тоже можно было добавить

oleksandrdubyna88:
Closes #294.

## What was asked

*"add reviewer нет клод кода"* — open **Add a reviewer** and Claude Code is not there; add it.

## What was actually wrong

The catalogue has had the entry all along, and there is only one Claude runtime (`RUNTIMES` in `models.ts`), so nothing was missing. **Two independent things hid it, and either one alone reads as "this product has no Claude Code":**

1. **The words were not in it, and the pick could not search the text that was.** The label was `Claude (a second one)`; `showQuickPick` was called with only `{ title, placeHolder }`, so VS Code filtered on the **label alone**. Typing *Claude Code* — the name the Consultant section itself uses (`consultSettings.ts`, `CALLER_KINDS`) — emptied the list.
2. **A preset whose id was already configured was dropped from the list, silently.** So an installation with a `claude` reviewer could not add a second one at all — no `claude` on haiku for the cheap passes beside a `claude` on opus for the hard rounds — and nothing said why the entry had gone. That is the same one-way door `VENDOR_PRESETS`' own docblock records for gemini.

The operator confirmed on 2026-09-15 that both should be fixed.

## What shipped

- The label is now **`Claude Code`**, and the options object carries **`matchOnDetail`** and **`matchOnDescription`**.
- The catalogue is offered **whole**. A preset whose id is taken comes with the next free one and an entry that says which.
- The offering decision **left the `vscode`-bound host** for three pure functions in `vendors.ts` — `freeVendorId`, `presetsOffered`, `reviewerPickItems` — so it is tested by calling it rather than by reading `panelProvider.ts` as text.
- The Team-server branch, the blank OpenAI-compatible entry, `askCustomEndpoint` and `saveVendor`'s duplicate refusal are unchanged in behaviour.

`research/module_extension.md`, `research/module_tests.md`, `CHANGELOG.md` and `README.md` updated; extension version to 0.47.0. The plan is promoted to `research/` with its deviations, in the same commit as its `git mv`.

## What the tests showed

Observed output, not "tests pass":

- **RED first.** Eight new assertions were watched failing against the unfixed code, naming the real symptoms: `actual: 'Claude (a second one)'` against `expected: /Claude Code/`; `actual: 'claude'` against `expected: 'claude-2'`; *"a preset already in the panel vanished from the list, which is the one-way door"*; and `actual: 1, expected: 8` when everything was configured. 8 failing, all of them new — nothing pre-existing broke.
- **GREEN.** Whole suite: **2944 tests, 2943 pass, 1 skipped, 0 fail** (baseline on `main`: 2932 / 2931 / 1 / 0).
- One assertion proved it has teeth during the round: changing the hint's casing turned it red (`/second/` against `A SECOND`) before the regex was made case-insensitive.
- `plan-lifecycle.mjs` clean, `pin-check.mjs` clean.

## The review gate

`coai`, both stages, on this branch.

- **Plan round** — `good_enough` (7 gating / threshold 6, all 3 reviewers answered; the plan stage allows one round). It caught a design bug **before any code**: the draft wrote `{ ...preset, id: offered.id }` where `offered` was the ARRAY — `undefined`, which would have written a row with no id *and* skipped `askCustomEndpoint` for the blank preset. One line, two defects. Also accepted: `matchOnDescription`, pinning the label test to `id === 'claude'`, and the pick-item seam.
- **Code round** — `proceed` (9 gating / threshold 5; all 12 reviewers answered, `Role2` not asked — its prompt `role2-general` has no text). 7 accepted, 11 rejected with reasons.

### What the gate found that changed the design

- **`(a second one)` left the label.** The plan kept it deliberately; the round overturned that and was right. Once a configured preset can be offered again, the pick also says *you already have one — this adds `claude-2`*, so the same words meant two different things two lines apart. The distinction moved into the hint, which `matchOnDetail` had just made searchable.
- **The id allocator reserves over the whole LIST.** Against the configured ids alone, a catalogue holding both `claude` and `claude-2` would resolve BOTH to `claude-2` whenever `claude` was configured. No preset has such an id today — a latent collision closed before it existed.
- **The structural assertion now pins the whole expression**, `existing` included. Matching `presetsOffered(VENDOR_PRESETS` alone stayed green if the host passed `new Set()`, which is precisely the defect that line exists to catch.

- **Four findings reported a crash when a Team server is picked** (`picked.offered!` on an entry whose `offered` is `undefined`). **Rejected — verified false**: `if (picked.server !== undefined) { … return; }` returns above that line, so a Team-server entry never reaches it. One of the findings says so itself mid-sentence, and gemini's own reliability pass recorded the branches as correctly discriminated. The assertion was nonetheless replaced by an explicit guard so nobody has to re-derive it from the control flow.
- **`research/architecture.md` should be updated** — rejected: no cross-module seam moved; `knowledge-base.md` scopes that file to exactly that case.
- **`npm test` is an unverified assumption** — rejected: the reviewer applied the .NET MTP rule to the Node half. `package.json` defines it and `ci.yml` runs it.
- **A quick-pick theme/truncation check** — rejected: nothing in a suite with no extension host can observe it. Its substantive half (filtering on `description`) was accepted as `matchOnDescription`.
- **Rename `second` to `isSecond`** — rejected: the neighbouring `Vendor` fields are `enabled`, `plan`, `code`.
- **Move `ReviewerPickItem` to its own module** — rejected until a second consumer exists (`reuse-first` §2).
- **A runtime type check on a `ReadonlySet`**, and **an O(N) concern in `freeVendorId`** — rejected: not a boundary, and a cost nobody has measured against a list a person types in by hand.

## Known gap

Nobody has driven the real quick pick. `matchOnDetail` / `matchOnDescription` are asserted as source text because this suite has no extension host — that gap is written into `research/module_tests.md` rather than implied.


oleksandrdubyna88:
**Automated-reviewer status, recorded rather than implied:** CodeRabbit did **not** review this pull request — it reported *"Review limit reached, next included review available in 34 minutes"* (free OSS tier). Its check nevertheless reports success, so the green row is not evidence that a review happened.

What did read the diff:

- **The `coai` gate's code round** — 12 reviewers across codex, gemini and local, verdict `proceed` (9 gating / threshold 5). 7 findings accepted and fixed, 11 rejected with reasons; all of it is in the description above.
- The extension suite: 2944 tests, 2943 pass, 1 skipped, 0 fail.

PR #305, opened 2026-09-15
fix(chat): a run with no spaces in it wraps instead of scrolling the page sideways

### LINKED ISSUE #299: Промт для ответа должен делать врам текст
Image
долежен быть врапа тексата, а не прокрутка вправо.

1. Публикуй драфт mcp-v0.23.0 прямо сейчас.
2. Релиз extension из текущего main не режем — чужую незавершенку в прод не тащим. 
[Вариант А, если #282 нужен срочно]: Сделай хотфикс extension-v0.46.1: ответвись от тега v0.46.0, сделай cherry-pick PR #282 и выпусти его.
[Вариант Б, если спешки нет]: Ждём отмашки от второй сессии по готовности их функционала, стабилизируем main и катим общий релиз.

oleksandrdubyna88:
Closes #299.

## What was asked

*"должен быть врап текста, а не прокрутка вправо"* — the prompt shown for an answer should wrap, not scroll sideways.

## What was wrong

`white-space: pre-wrap` wraps only where a break opportunity already exists — in practice, a space. A Windows path, a URL, a stack frame or a base64 blob has none, so it cannot wrap and it overflows. Two rules were in that state: **`.msg .what`** (the body of every message; `.msg.you .what` adds the `pre-wrap`) and **`.passage`** (the captured code a question is asked about).

The spill becomes a **scrollbar** rather than an overhang because of the layout: `body` is `overflow: hidden` so the page itself cannot scroll, and `#scroll` sets only `overflow-y: auto` — and per CSS, `overflow-x: visible` **computes to `auto`** when the other axis is not `visible`.

The question is the widest thing on the page because a `you` bubble is not what the person typed: it is the composed prompt — instruction, language line, fence, and the whole captured passage verbatim, which is usually code.

## What shipped

```css
.msg .what  { …; overflow-wrap: anywhere; }
.passage    { …; overflow-wrap: anywhere; }
.msg .what pre, .msg .what table { overflow-wrap: normal; }
```

**`anywhere`, not `break-word`** — although three siblings on this page (`.askedText`, `#say`, `#ghost`) use `break-word`. The plan round surfaced the one way they differ: `break-word` does not shrink a box's intrinsic **min-content** width and `anywhere` does, so `break-word` leaves the scrollbar wherever a box is sized *by* its content. That is **not** the case in this tree — `#scroll` is a flex item of a *column* container, so its automatic minimum is on height (which is why the neighbouring comment sets `min-height: 0`), and `.msg` is a plain block — so `break-word` would have worked. `anywhere` was taken anyway: it is free, `panelView.ts` already settled on it for this same symptom, and it ends the question instead of leaving it to be re-derived.

**The third rule is the one the plan did not have.** `overflow-wrap` is inherited, and two boxes inside `.msg .what` scroll on purpose. The `pre` would have been safe *by accident* — `white-space: pre` leaves the property nothing to act on — and safe by accident stops being safe the day somebody makes it `pre-wrap`. The **table** is the real one: it is `display: block; overflow-x: auto` and its cells *do* wrap, so the inherited wrap would have broken a long token in a cell, re-flowed the columns and quietly removed the scroll the rule was written for.

`.msg .what pre` now also states `white-space: pre` rather than leaning on the UA default, because with the wrap inherited the whole decision rests on that one value.

## What the tests showed

- **RED first.** All three assertions were watched failing: *"a long path in a question cannot break, so it widens the bubble and the conversation scrolls sideways"*; the same for the passage; and *"there is no rule for `.msg .what pre, .msg .what table`"*.
- **GREEN.** Whole suite: **2947 tests, 2946 pass, 1 skipped, 0 fail**.
- `plan-lifecycle.mjs` and `pin-check.mjs` clean.
- One build break on the way, worth recording: a backtick inside a CSS comment ends the stylesheet's template literal. Caught by `tsc`, and the repo already has a standing guard for it (`chatPage.test.ts`, *"the page module carries no backtick inside its own template literals"*).

## The review gate

`coai`, both stages.

- **Plan round** — `good_enough` (9 gating / threshold 6, all 3 reviewers). It changed the fix: `anywhere` over `break-word`, the table exclusion, and an honest relabelling of the third test.
- **Code round** — `proceed` (10 gating / threshold 5; all 12 reviewers answered, `Role2` not asked — its prompt `role2-general` has no text). 5 accepted, 11 rejected with reasons.

### What the code round changed

- **A test comment was lying.** It said *"A REGRESSION GUARD, green before this change and after it"* — it was not; its first assertion went red because the exclusion it asserts is new code. Fixed to say which half is which.
- **The table assertions now pin the whole condition** — `display: block` and `overflow-x: auto`, not just the `overflow-wrap` override. Without that, deleting the table's scroll left the test green while the exclusion protected a box that no longer scrolled.
- **`white-space: pre` is stated explicitly** on the code block, so the test can hold it.
- The plan's Definition of Done is ticked; unticked boxes in `research/` read as verification nobody did.

- **Three findings proposed adding `white-space: pre-wrap` to `.msg .what pre`, and one `word-break: break-all`.** Rejected: each would make code blocks break mid-token, which is the exact decision the box exists to hold — and two of the findings said so themselves one sentence earlier. The real point underneath (the decision rested on a UA default) was taken the correct way: explicit `white-space: pre`.
- **"Add a reset rule for `pre` and `table`"** — rejected: that rule *is* the change under review.
- **Two findings on backtick risk** — rejected: no backtick is present (it would not compile), and a standing test already scans for them, escaped ones included. The meta-comment that prompted them was removed anyway.
- **Three findings asking for a rendered/browser test** of actual wrapping — rejected: this suite has no layout engine, so such a test would assert numbers the DOM shim itself set and would look like evidence while proving less than the declaration. Written into `research/module_tests.md` under *What this does NOT prove*.
- **A semantic scroll-box class** for future scrollers — rejected until a third one exists.

## Known gap

No test here proves that text wraps — only that the declarations are in the stylesheet. Recorded in `research/module_tests.md`.


oleksandrdubyna88:
**Automated-reviewer status, recorded rather than implied:** CodeRabbit did **not** review this pull request — it reported *"Review limit reached, next included review available in 12 minutes"* (free OSS tier). Its check nevertheless reports success, so the green row is not evidence that a review happened.

What did read the diff:

- **The `coai` gate's code round** — 12 reviewers across codex, gemini and local, verdict `proceed` (10 gating / threshold 5). 5 findings accepted and fixed, 11 rejected with reasons; all of it is in the description above.
- The extension suite: 2947 tests, 2946 pass, 1 skipped, 0 fail.

PR #308, opened 2026-09-15
feat(panel): each Consultant row is a framed group in its client colour

### LINKED ISSUE #291: выделить цветами как в ревьюверах
Image
вот так сейчас выглядят секции в консультант.

Image
а так в ревьюверах. 
1. то что относится к клоду (кодексу, джемини, другой килент) - взять в рамку (группа)
2. слева полоску цвета задать (причем цвет долждне совпадать с другими секциями в зависимости от провайдера)
3.

oleksandrdubyna88:
Closes #291.

## What was asked

The operator, comparing **Consultant** with **Reviewers**:

> (1) what belongs to claude (codex, gemini, another client) should be taken into a frame — a group; (2) give it a coloured stripe on the left, and the colour must match the one that provider has in the other sections.

## What was wrong

**REVIEWERS** draws one bordered card per vendor with a 3px left edge in that vendor's colour, and the same colour follows it into *Active rounds* and the rounds log — so a vendor can be walked from its settings to its running round without reading.

**CONSULTANT** drew four flat `.field` blocks. No frame, nothing grouping a caller's controls, no colour anywhere. `.consultant-row` existed as a **data** hook with no CSS rule in the panel at all.

## What shipped

Each caller row is now a framed box with a coloured left edge, and the colour comes from **the same allocator** — `vendorPalette(state.vendors.map(v => v.id))`, the expression `reviewersBody` already uses. That is what makes the cross-view promise true by construction: a palette built from the four caller kinds would be a second list and a second assignment, which is the defect `vendorColour.ts` exists to prevent.

- **`claude`, `codex` and `gemini` are ANCHORED** in the palette, so they answer with the colour they already have whether or not a reviewer of that name is configured. The palette is therefore passed even when `state.vendors` is empty — an anchor exists for exactly that case.
- **`other` takes a neutral edge** (`var(--vscode-widget-border)`). *Another client* is not a vendor and must not look like one; and being unanchored, a hashed colour for it could permanently collide with a configured reviewer's. Anchoring it was rejected — a sixth reserved slot would **move the colours of currently unanchored reviewers**.
- The frame is modelled on `.role`, not `.vendor`: its neighbours here are fields in a section, not cards in a list. Joining the selectors was rejected — it would re-space these rows on any future change to the reviewer card, and pull in four descendant rules written for controls a consultant row does not have.

## What the tests showed

- Whole suite: **2960 tests, 2959 pass, 1 skipped, 0 fail**.
- `plan-lifecycle.mjs` and `pin-check.mjs` clean.

**RED was obtained by reverting, not before the code — and that is recorded as a deviation rather than papered over.** The tests and the implementation were written together and went green first time. Corrected the way the rule prescribes for a fix that landed first:

1. the stripe removed → *"claude's row carries no edge colour of its own"*, *"codex is a different colour in Consultant than on its reviewer card"*;
2. restored, then the **wiring** deleted → **only** the integration assertion went red while every `callerColour` test stayed green;
3. restored, then the **CSS rule** deleted → *"the `.consultant-row` rule is gone, so the rows are unframed whatever colour they carry"*.

Each of those three reverts targets a different production line, and each reddened a different test. That is the evidence that the suite guards the feature rather than the function beside it.

## The review gate

`coai`, both stages.

- **Plan round** — `good_enough` (8 gating / threshold 6, all 3 reviewers). 6 accepted, 5 rejected.
- **Code round** — `proceed` (10 gating / threshold 5; all 12 reviewers answered, `Role2` not asked — its prompt `role2-general` has no text). 6 accepted, 13 rejected with reasons.

### What the gate caught that was real

- **The framing test did not pin the frame.** `consultantBody` emits no stylesheet, so asserting each row's inline `border-left-color` stayed green after `.consultant-row` lost its rule — and the rows would go back to being flat blocks with a coloured line, *which is the reported defect*. Now asserted where the page and its stylesheet are the same string.
- **The cross-view assertion was not scoped to the codex card.** It looked for *any* `.vendor` div wearing codex's colour, so a page that swapped two vendors' colours would satisfy it twice over while the codex-to-codex equality it is named for was false. The card is now found by the id its own label carries.
- **The plan round caught that the unit tests guarded nothing on their own** — `callerColour` can be perfect while `panelHtml` never hands the palette over. That is where the integration assertion came from.
- `ConsultantViewState.colour` → `palette`: it holds a lookup **function**, and a singular `colour` is what the reviewer card's `CardContext` calls a string.
- Two corrections to the plan document: the build order still claimed the tests were watched failing as step 1, and the DoD ticked "four framed rows" without saying their **visibility** is unproven.

- **A CSS-injection finding (Blocking).** Traced and rejected: `vendorPalette` returns one of thirteen fixed internal constants, and the vendor id is the lookup **key** — it never reaches the output, so a hostile id yields a different fixed constant, not a hostile string. There is no quote or semicolon to escape, the identical unescaped idiom already ships on the reviewer card, and `escapeHtml` is an HTML escaper, wrong for a CSS value.
- **"The rows lack a complete frame"** — the rule it describes is not the rule that was written, which carries `border: 1px solid` *and* `border-radius`. That two reviewers reached opposite conclusions about it is precisely why the rule is now asserted rather than only written.
- **Four findings asking the integration check to run a bundled page** — there is no CSS engine anywhere here, and no sidebar page harness at all (`bundledPage.test.ts` bundles `roundsLog.ts`). A DOM-shim assertion would read values the shim was handed. Written into `research/module_tests.md` under *What this does NOT prove*.
- **A cluster about `undefined`/non-function palettes** — the types and the fixed palette both preclude them, and `coding-style.md` scopes validation to system boundaries, which a same-package call is not.

## Open PR count

This leaves **4 open on the repository** — two Dependabot (#239, #240) and one from a parallel session (#307), plus this. Only this one is mine; #306 was merged immediately before opening it, so my own queue is one.

## Known gap

That the frame is **visible** is not proven — only that it is declared. Nothing in this repository has a layout engine.


oleksandrdubyna88:
**Automated-reviewer status, recorded rather than implied:** CodeRabbit did **not** review this pull request — it reported *"Review limit reached, next included review available in its stated reset"* (free OSS tier). Its check nevertheless reports success, so the green row is not evidence that a review happened.

What did read the diff:

- **The `coai` gate's code round** — 12 reviewers across codex, gemini and local, verdict `proceed` (10 gating / threshold 5). 6 findings accepted and fixed, 13 rejected with reasons; all of it is in the description above.
- The extension suite: 2960 tests, 2959 pass, 1 skipped, 0 fail.

PR #312, opened 2026-09-16
feat(spending): a chat row can be cleared from the chart, as a reviewer row can

### LINKED ISSUE #298: кнопка крестик (удалить) на айтемах чата
Image
аналогично Reviewers должна быть возможность удалить и в чат айтемах , с той же логикой (удалил, если появилась - продолжнается с чистого листа)

oleksandrdubyna88:
Closes #298.

## What a person sees

The spending chart's **reviewers** half has had a ✕ on every row since it shipped — press it and that provider's past spending stops counting. The **chat** half had nothing. A row from a model you tried once and dropped sat in the chart forever, and the only way to clear it was to not look at the chart.

Now each chat row carries the same ✕, and it forgets the same way.

## Why a chat row needed more than the reviewer one

A reviewer row is a **vendor**. A chat row is a **vendor AND a model** — `codex / gpt-5.4` and `codex / gpt-5.6-sol` are two rows, and one id cannot name the pair. So:

- the button carries `data-model` beside `data-id`,
- the page's delegated `[data-command]` handler posts **both** (it posted only `id` before — the markup was right and the page dropped the field, which is the defect the gate's "run the page" findings were pointing at),
- the watermark map is keyed by the pair: `coai.chatUsageForgottenBefore`, separate from the reviewers' `coai.usageForgottenBefore`.

## How forgetting works

The same pattern the reviewers half already uses, unchanged: nothing is deleted. A watermark timestamp goes into `globalState`, and `rememberedChat` applies it **on read**, to both ledgers — turns and doors. No file is ever rewritten, so a forget is reversible by anyone who clears the key and costs nothing at write time.

## Tests

`src_vs_code/src/test/forgetAChatRow.test.ts` — 8 tests, and the middle one **executes the page's own script** rather than reading it: the listener the page registers on `document` is captured and called with an event whose target answers `closest` the way a real one would. That is the only test in this file that could have caught the dropped field, and every other assertion over the rendered HTML was green while the defect was live.

Also covered: the escaping of a hostile model id (a quote in `data-model` would forge a second attribute and forget a pair nobody pointed at), the row with nothing chosen getting **no** control (a button that cannot act is worse than none), and a source assertion that both ledgers actually reach the page through the filter — `panelProvider.ts` needs a VS Code host this suite has none of.

Suite: **2984 tests, 2983 pass, 1 skipped, 0 fail.** `plan-lifecycle` clean, `pin-check` OK.

## The coai gate

Both stages complete against all 12 reviewers.

- **Plan** — `good_enough` (the plan stage's round budget is 1, so that is the pass), 13 gating findings of 6 raised.
- **Code** — `proceed`, 15 gating of 5. **6 accepted, 21 rejected with reasons.**

Four rejections are worth naming, because they were checked rather than argued: four findings claimed `chatLedgers()` runs once at startup and would serve a stale ledger forever. It is called inside `usageTab()`, which runs per render — verified at the call site before rejecting.

Two accepted findings changed this branch after the code round: the fixture's hour moved to local **noon** (the only hour no inhabited offset can move out of its own day — the window is built from the local calendar day), and the running-page helper returns a pair instead of mutating the function it captured.

## Docs

`research/PLAN_a_chat_ledger_row_you_can_clear.md` promoted with its deviations and a ticked DoD; `research/README.md`, `research/module_extension.md` and `research/module_tests.md` updated.


oleksandrdubyna88:
**CodeRabbit did not review this PR** — its check reports *"Review rate limited"* (the free OSS tier's budget for this repository is spent). Recorded here rather than re-triggered: asking for a review consumes the reset window and makes the wait longer, so the gap is noted instead.

What did run: CodeQL (actions, javascript-typescript), SonarCloud, the extension typecheck/test/package job and the family checks — all green. The change also went through the coai gate on both stages, with 27 findings resolved (6 accepted, 21 rejected with reasons).

PR #316, opened 2026-09-16
feat(roles): the Edit-roles page is three tabs, what you added is green, and the colours match the menu

### LINKED ISSUE #293: пустить секции в Edit roles в табы на странице
Image
тут сильно много всего. кажду секцию пустить в  табы

Image
при нажатии на адд новый промт - 
он должен быть в зеленой рамочке - иначе не понятно куда он добавитлся, что его можно редактировать. и он зеленый всегда остается (то что можно редактировать)

Image

Image
линии цвета слева должны по цвету совпадать с тем, что уже есть в меню. сейчас все синии в едит

oleksandrdubyna88:
Closes #293.

## What a person sees

Three things on the **Edit roles** page, from the three pictures on the issue.

### 1. It is three tabs, and the tab you are on survives

The page was one column of everything: seven roles, each with up to six prompt boxes ten rows high, under three headings on one scroll. *Plan review* · *Code review* · *Document review* are tabs now — same sections, same order, same content.

**The tab is held by the host, and that is the part that is not decoration.** `apply` in `rolesPanel.ts` returns whether the page must be redrawn, and a redraw replaces `panel.webview.html` wholesale — a new document with no memory of anything. Every shape-changing action redraws: add a prompt, add a role, a switch, a stage, a remove, a restore. A page-local tab would mean pressing **Add a prompt** on an architecture role and being thrown back to the plan tab, with the prompt you just added on a tab you can no longer see — the complaint in the issue, made worse.

So the page switches instantly on the press **and** posts `{ type: 'tab', id }`; the host keeps it and hands it back on every render. Adding a *role* moves you to *Code review*, because that is where a new role appears; adding a *prompt* leaves you exactly where you are.

The whole rule is **`nextTab(current, command)`** — a pure exported function beside the page rather than a rule buried in the host, so a test executes it instead of reading source text. `.agents/PROJECT.md:78` refuses a new behavioural assertion over page source text, and the plan round named this test in particular.

### 2. A prompt you added is framed in green, and stays framed

On the same condition that already chose **Remove** over **Restore**: `isShippedPrompt` is false. It is not a highlight of the last thing added — it is what the block *is*, the one whose label, purpose and text are all yours to edit. A shipped prompt you have rewritten is still shipped: it keeps its grey edge and its **Restore**.

### 3. One palette, not two

`ROLE_TONE` and the `--tone-*` block moved out of `panelView.ts` into **`roleTone.ts`**, which both views now import. The sidebar had coloured each role individually since the settings panel was written; this page gave every role the same blue. A role you added takes its stage's tone, which means it shares a colour with a shipped one — matching the sidebar means matching it where it is arbitrary too.

## The coai gate

Both stages, all 12 reviewers. **Plan**: `proceed` on the first round, 11 findings, 9 accepted / 2 rejected. **Code**: `good_enough`, 31 findings, 18 gating, **9 accepted / 22 rejected with reasons**.

**The code round found a defect that shipped in the first commit and that none of this branch's own tests could see.** The shared palette was declared *above* the base `.role` rule. `.role` and `.role-arch` have equal specificity and `.role` sets the whole `border-left` shorthand, so the later rule wins — the palette never applied, and **every edge on the page would have rendered in one colour**, which is the state the issue reports with the classes added. Two reviewers found it independently; codex named why the tests missed it (the cross-renderer test compares class *names*, and the names were right). Measured before accepting it: `.role-plan` at index 1687 of the stylesheet, `.role` at 2797. The palette moved below the base rule and a new test reads the **order** of the two rules, because there is no CSS engine in this suite to ask for a computed colour. It was correct in `panelView.ts` and wrong here — which is what a second consumer of an extracted block is for.

Also taken: **Add a role** moved inside the *Code review* section (it sat outside all three and quietly created something on another tab); the running-page test now builds its nodes out of the rendered HTML instead of from `ROLE_TABS`, so a page drawing `data-panel` while its script queries `data-section` can no longer pass; the cross-renderer list comes from `BUILTIN_ROLES`; and the tabs gained `role="tablist"`/`"tab"`/`"tabpanel"`, `aria-selected` and `aria-controls`, kept in step by the script.

Four rejections worth naming, each checked rather than argued:

- *"The tab click handler is missing its `return` and falls through to the role switch"* (Blocking) — it is there, on the line after the post; the branch is also first in the handler and no later selector matches a tab button.
- *"The harness's `new Function` is a security defect"* (Blocking) — executing the shipped script is what `.agents/PROJECT.md` requires, the input is a fixture the test writes two lines above, and the pattern predates this change.
- *"An unknown tab id should be refused naming the legal values"* — a webview message is not a name a person typed, and `roundsLogMessages.ts` settles this in writing: a retained webview can be older or newer than the extension, so an unknown word is ignored. Both halves normalise, so exactly one tab is always open.
- *"`module_tests.md` must mark every flow covered or not covered"* — that file contains those words zero times; it is prose, and the convention quoted belongs to another repository.

## Tests

**Written first and watched red**, with the failures naming the real symptom:

- `the page does not offer the three sections as tabs`
- `pressing a tab told the host nothing, so the next repaint will undo it`
- `PlanCritique is toned arch, and no .role-arch rule colours it`
- and, after the code round, `.role-plan is declared before .role, so the base border-left overrides it`

`tsc` clean on every run before any suite number was read. Final: **3008 tests, 3007 pass, 1 skipped, 0 fail.** `plan-lifecycle` clean, `pin-check` OK.

New: `roleTone.test.ts`, `editRolesInTabs.test.ts`. The DOM shim was **extracted** into `rolesPageHarness.ts` rather than copied — `rolesPageScript.test.ts` now uses it too. One existing test was rewritten, not weakened: `rolesPage.test.ts`'s "the plan stage and the code stage are drawn apart" located the groups by their `<h2>` headings, which are now tabs; it asserts the same guarantee against `data-section`, with one more assertion than before.

## Docs

Plan promoted to `research/PLAN_edit_roles_in_tabs.md` with both rounds recorded, deviations, and a ticked DoD. `research/README.md`, `research/module_extension.md` and `research/module_tests.md` updated.

## One question, not blocking

The rounds log page's tab strip has no ARIA either. This PR did not touch it — rewriting neighbouring code uninvited turns a small change into a large diff — but it is the same three attributes. Worth a follow-up?


oleksandrdubyna88:
**CodeRabbit did not review this PR** — the check reports *"Review rate limited"* and there are zero review comments on the diff. The row is green because the check passes on a rate limit, not because anything was read. Recorded rather than re-triggered: asking for a review consumes the reset window and makes the wait longer.

What did read this change: the coai gate, both stages, all 12 reviewers — 42 findings across the two rounds, 18 accepted and 24 rejected with reasons. The code round caught a real defect the branch's own tests could not see (the shared palette declared above the base `.role` rule, so every edge would have rendered in one colour); it is fixed in `14b0e40c` with a test that reads the cascade order.

Machine checks that did run: CodeQL (actions, javascript-typescript), SonarCloud, `extension · typecheck · test · package`, `build · test · family checks`, `submit-nuget`.

PR #335, opened 2026-09-16
feat(panel): the Claude models are asked, rather than listed

### LINKED ISSUE #301: для консультатнта при выборе клода нет фабл
Image
должно брать реалтайм текущие доступные модели из соотв кли (кодекс, клод, джеминай ). 
плюс по умолчанию берем самую высокую версию (н)апример фабл 5 и 5.1 ) если доступно. 
плюс для консультанта должно быть возможность выбрать в том числе локальную модель

oleksandrdubyna88:
Closes #301.

Claude was the only runtime whose model list was hard-coded. codex reads a cache file, antigravity answers `agy models`, a local engine answers its own catalogue — and the curated Claude array rested on a theory that breaks the moment a new family appears. **`fable` is a new family, it IS reachable on this machine, and it had never been offered.** The caption underneath made it worse by stating as fact something nobody had checked: *aliases the Claude CLI resolves to the latest of each family*.

## Two measurements decided the design, and both refuted the obvious approach

**`/v1/models` cannot be called from here.** No Anthropic key exists anywhere this extension can read — not the settings, not the CredsForDevs entries opened to it, not the host's environment — and the CLI holds its own credential without handing it out.

**And the naive CLI probe is also wrong.** `claude --model <anything> -p hi` exits **0** and answers from the DEFAULT model, silently ignoring a name it has never heard of. A probe built on exit codes would have reported every candidate available, which is worse than the array it replaces. What does answer is `--output-format json`: the reply carries a `modelUsage` object keyed by the model that actually ran. So the question asked is not *did the call succeed* but *is the family that answered the family that was asked for*.

| module | job |
|---|---|
| `claudeModels.ts` | the decisions, all pure: `familyOf`, `answeredAsAsked`, `modelThatAnswered`, `stillGood`, `claudeModels`, `claudeNote` |
| `claudeProbe.ts` | the run: one small request per candidate, sequentially, through `versionProbe.ts`'s `capture` |
| `claudeProbeFile.ts` | `claude-models.json` in the data directory — rebuilt entry by entry on read, written atomically |
| `claudeCli.ts` | which binary to ask, and whether to ask at all |
| `lookingSpinner.ts` | the turning ring both surfaces wear while the probe is out |

## A discovery that cannot run never subtracts

The three failure states are all ones this installation is really in from time to time: an allowance that is spent (#165 just measured four reviewers meeting exactly that), a CLI that is not installed, and a first run. In every one of them the curated list is offered in full — `fable` included, which is why that fix ships on its own line and does not depend on the probe.

What changes is the **label**: a verified entry names the concrete id the CLI resolved (`sonnet — claude-sonnet-5`), an unverified one says *not asked yet*. Offering an unverified `fable` unlabelled would mean a round reviewed by whatever the default is, with nothing on screen saying so.

Three rules protect that promise, and each came from a reviewer:

- a run that could not ask every candidate **merges** into what was held rather than replacing it;
- a run that confirmed **nothing** — what a spent allowance produces, every candidate answered and none verified — never displaces a verdict that was confirmed;
- nothing is carried across a CLI version change at all.

## The panel says it is looking

`refreshClaudeProbe` is shaped like `refreshProviders` beside it: a render starts it and never awaits it. It is **edge-triggered on the executable**, because the first draft started a refresh on every render and repainted from its `finally` whatever it found — a repaint that started another refresh, which spawned another `--version`, for ever. A failure is **dated rather than latched**, so a timeout or a spent allowance can be retried after ten minutes instead of locking the dropdown until the editor restarts.

## And the consultant's local dropdown finally asks an engine

It passed `undefined` as `modelsFor`'s `localEngine`, so the list was empty and a saved model was labelled **"NOT on this engine any more"** — false, and alarming, because nothing had asked. The panel probes consultant **endpoints** now, in a second map keyed by endpoint rather than vendor id, because since story C5 that section holds no reviewer rows to borrow one from. `modelsFor` gained the third local state that goes with it: *not asked yet* is a different sentence from *did not answer*.

## The live contract check, and what it caught within a minute

`npm run test:claude-models` drives the **real** CLI and asserts that `modelUsage` is keyed by exactly one model, that a known family answers as itself, and that an unknown name still exits 0. It is not in CI — each candidate is a billed request — and it is the only thing that can notice the CLI changing its JSON shape, which would otherwise leave the whole suite green with every candidate quietly unverified. Run 2026-09-16 against CLI 2.1.258: **the contract holds.**

It earned itself immediately. `capture` left the child's stdin open, so the Claude CLI waited three seconds for input on every call — twelve seconds per probe, spent waiting for a pipe `capture` has no way to write to. It sends EOF now, pinned by a `processLauncher.test.ts` case that goes red in five seconds against the old code.

## The gate

| round | verdict | findings | accepted | rejected |
|---|---|---|---|---|
| plan | `good_enough` | 17 | 16 | 1 |
| code 1 | `revise` | 41 | 28 | 13 |
| code 2 | `proceed` | 9 | 7 | 2 |

The four that were defects a person would have met: the render loop; a partial run subtracting; the probe asking whatever PATH answered when only a consultant named its CLI; and a stored `verified` believed instead of re-derived from the `answered` beside it.

Both rejections in round 2 are recorded with reasons — one asked for the give-up count that round 1 had removed for cause, and one asked for a check it states three times is already correct.

## Test plan

- [x] `cd src_vs_code && rm -rf out && npm run compile && node scripts/run-tests.mjs` — **3145 tests, 3144 pass, 0 fail, 1 skipped**, on the branch rebased onto `origin/main` with the conventions submodule at its committed pin
- [x] `npm run bundle` clean
- [x] `node .agents/conventions/tools/plan-lifecycle.mjs` clean
- [x] `npm run test:claude-models` against the installed CLI 2.1.258 — the contract holds
- [x] **Eleven teeth checks**, each compiled and each reddening only the test named for it: the row that passes no engine; an unasked engine reported as a refusal; a row taking whichever engine is first; a probe ignoring a configured CLI path; one failed candidate abandoning the run; a partial run replacing what was known; a stored verdict believed; the file parsed unbounded; an unconfirmed answer displacing a confirmed one; the record forgetting its binary; a failed probe latched for the session; a stale verdict reaching the chat
- [ ] Not covered: no extension host in CI, so the probe running inside a real window is verified by installing the build

## Docs

`research/module_extension.md` (*The Claude list is ASKED, not listed*), `research/module_tests.md` (the flow, and what the hand-run script does and does not prove), `research/architecture.md` (a discovery whose peer is a CLI, with the two-sided table), and the plan promoted to `research/` with what shipped differently.

## Summary by CodeRabbit

* **New Features**
  * Claude model availability is now checked through the configured CLI, with results cached for one week.
  * Model dropdowns show verified model information and indicate when models have not yet been checked.
  * Consultant local-model lists are discovered separately for each endpoint.
  * Added a visible loading indicator while model availability is being checked.
  * Added support for the Fable Claude model family.

* **Bug Fixes**
  * Improved handling of unavailable or unresponsive model and consultant endpoints.
  * CLI checks now complete without unnecessary waiting for input.


oleksandrdubyna88:
@coderabbitai review

oleksandrdubyna88:
All six applied — thank you, and one of them was the best finding of the whole review.

**The test with no teeth.** You were right, and the break proves it: the give-up counter resets after a candidate that answered, and my case failed only the *first* candidate, so one failure never reaches the bound whether or not the reset is there. Deleting `couldNotRun = 0` left it green. The new case separates the failures — `haiku: -1`, `sonnet` answering, `opus: -1`, `fable` answering — and deleting the reset now reddens it.

**Executable identity through `modelsFor`.** The gap was real and wider than the chat: `claudeExecutableFor` picks one binary for the machine, so a *second* Claude reviewer row naming another would have worn the first one's labels too. `probeFor(probe, executable)` withholds a record taken from another installation, and every caller now names the binary it will run — the reviewer card its row's, the consultant row its own, the chat picker the row it resolved. Legacy behaviour is explicit, as you asked: a record with no `executable`, and a caller holding none, are both left alone, because neither has made a claim.

**Bounding before the parse.** Moved onto the bytes, before the decode — `LONGEST_FILE = 64 KiB`, two orders of margin over anything the writer produces. Each `asked` and `answered` is bounded too, and an entry carrying a runaway string is dropped on its own rather than taking the others with it.

**"Not asked" against "asked, confirmed nothing".** Its own sentence now, with the date: *the Claude CLI was asked 2026-09-16 and confirmed no family — the curated list stands.* That is what a spent allowance looks like, and it was the one fact a person could act on.

**The shared Windows-side memo.** Correct, and it was introduced in this PR: moving the consultant probe off the render path made it detached, so clearing one shared field at its own start and end reached into the awaited reviewer pass. Each pass holds its own memo now, and `probeConsultEngines` no longer touches `windowsSideThisPass` at all.

**The rot guard.** Added, and it checks the imported symbols against the declarations in `claudeModels.ts` rather than against a list kept in the test — a list would be a third place to update. Plus a check that `package.json` still runs it, since a script nothing runs is a script nobody runs.

3169 tests, 3168 pass, 1 skipped, from a cleaned `out/`. Five teeth checks on these fixes, each compiled and each reddening only the test named for it.

PR #337, opened 2026-09-17
fix(bugs): the copy undid the setgid on the line after it was granted

oleksandrdubyna88:
## The defect

`bugs-v0.1.0` is deployed and serving. Checking *how* it reaches its own binary turned up this:

```
/opt/coai-bugs/releases            coai-bugs-deploy:coai-bugs        drwxr-s---  2750
/opt/coai-bugs/releases/0.1.0-...  coai-bugs-deploy:coai-bugs-deploy drwxr-xr-x   755
```

The parent is setgid, so a release directory created under it inherits `coai-bugs` and the service reads its binary **by group**. The child did not inherit it. The service was reaching the executable through the **world bits** — which is precisely what the setgid bit was added to prevent, so the earlier commit that added it did not do what its message claimed.

The cause is the line after the directory is created:

```sh
cp -a "$found/." "$RELEASE/"
```

The trailing `/.` makes `cp` apply the **source** directory's mode, group and timestamps to the destination. The archive's directory attributes landed on `$RELEASE`, wiping the inherited group one line after it was granted. It also explains a detail that looked like nothing: the release directory's mtime was the moment the *artefact was built*, not the moment it was installed.

The deployment worked, for a reason nobody chose. It would have stopped working the day an archive was packed with a directory mode of `0750`.

## Measured, not argued

A claim about setgid is worth checking, so it was checked in a scratch directory on the host itself:

| step | the directory read |
|---|---|
| plain `mkdir` under the 2750 parent | `coai-bugs 2755` |
| after `cp -a source/. copied/` | `root 755` |
| after `chgrp --reference` + `chmod 0750` | `coai-bugs 750` |

## The fix

The group is put back explicitly after **either** install path:

```sh
chgrp -R --reference="$RELEASES" "$RELEASE"
chmod 0750 "$RELEASE"
```

- `--reference` takes the group from the parent instead of naming it, so renaming the service account cannot silently break this.
- `-R` because the files need group access too — the binary to execute, `e_sqlite3` to `dlopen`.
- `chmod 0750` removes the world access the copy granted, so the group is doing the work rather than standing beside a wider permission.

Both halves are needed and the notes now say so: the setgid bit for the `dotnet publish` path, this line for the archive path. The notes previously credited the bit alone.

## Verification

- `release.sh` parses (`bash -n`); no carriage returns.
- The **live** release was corrected the same way and checked by hand: with the world bits gone, `sudo -u coai-bugs test -x` on the binary and `test -r` on `libe_sqlite3.so` both pass, the unit restarted, and `/health` answers `{"status":"ok"}` on loopback and from the public internet.

## Unrelated, and done on the host rather than here

The unit was **`disabled`** — installed and running, but it would not have come back after a reboot. It is `enabled` now. Nothing in the repository was wrong; the one-time setup notes say to install the unit and never say to enable it, which is worth a follow-up if this recurs on the next host.


oleksandrdubyna88:
Added the enable gap to this PR rather than leaving it as a follow-up — it is the same file and the same class of defect: a state nothing reports, which this deployment was already in.

`systemctl is-active` answers `active` whether or not the unit is enabled, and so does the deploy's canary, which asks whether the service answers rather than whether it will exist after a reboot. `systemctl is-enabled` is the only thing that answers it.

